PR #19, opened 2016-05-04
Refactoring time service and increasing test coverage

jarlah:
1. Moved all business logic from controller to service
   - Easier to test individual parts of the business logic
2. MusitJodaFilter and MusitTime is now sealed traits
   - Compiler will now nag if matching on those types does not include all implementations
3. Adding tests for controller and service with test coverage now on 87%

---

This change is [Reviewable](https://reviewable.io/reviews/musit-norway/musit/19)


jarlah:
No worries about working overtime. I just checked in what I worked on earlier today. And added an extra test while poking at it

jarlah:
These changes needs a little bit more love tomorrow.

PR #106, opened 2016-08-11
MUSARK-219

kpmeen:
**NOTE**: This PR is quite large. It refactors almost all service_\* (except service_event), to use the simplest form of the Play! default dependency injection framework (Google Guice).

There are a few reasons why this is good:
- First of all the code is now easier to test in isolation. You do not need to initialise everything to be able to test a focused part of the code.
- No more use of deprecated Play API's (these **will** disappear from the framework in the not so distant future... for good reasons)
- Instances of *Dao and *Service are now handled properly by the Play! application lifecycle mechanisms. (less resource leaks etc...).
#### What has been done:
- All *Dao implementations have been changed from object -> singleton class
- All *Service implementations have been changed from trait -> class (with appropriate Dao and \* Service dependencies injected in the constructor)
- All *Resource implementations now get their Service dependencies injected in the constructor.

This exercise also uncovered and took care of a few circular dependencies and use of code that bypassed the layer structure and responsibilities.

Last but not least, the service_\* applications that have been refactored have a much nicer memory footprint in SBT causing much less OutOfMemoryExceptions (i.e. resource/memory leaks).
#### What has _NOT_ been changed:
- `service_event` has not yet been refactored. This is because the implementation is quite complex. Especially they service and domain layers need to be looked at in detail.
- `common` has an implementation of (at least) `LinkDao` that should be refactored.


jarlah:
Har sett grovt igjennom endringene og kan ikke se noe med det. LGTM

PR #108, opened 2016-08-15
MUSARK-228

rituvesh:
Updated the URL's and added them on new observation & new control buttons.


jarlah:
Great job! :) but when i look at the urls you defined, i proposed a prefix which you omitted "node/". However, this is something that is connected to the "magasin" domain. Because of this i feel it more natural to have urls like this: "magasin/:id/obersvation/add", "magasin/:id/observation/:obsId" and etc. We can discuss this in plenum tomorrow after standup.

jarlah:
closing. transferred to new repo

PR #124, opened 2016-09-02
Fixing stacktrace in console output due to missing dependency in common.

kpmeen:
The common module did not include a dependency to `play-logging`. This caused the ugly stack trace in the console every time you ran a build/tests.
Also changed log-level of slick jdbc from DEBUG to WARN in tests to reduce verbosity. If you want that verbosity back , you should change the appropriate logback.xml locally. And not include the change in any commit.

Also did some other minor cleanup in the build scripts.
BaseEventDTOTest.scala seemed not have been formatted by scalariform before it was pushed at some point.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,14 +54,25 @@ val noPublish = Seq(
   publishLocal := {}
 )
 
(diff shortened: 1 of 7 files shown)


jarlah:
LGTM

PR #128, opened 2016-09-09
removed unnecessary and unused plugins

kpmeen:
These just waste unnecessary amounts of bandwidth and disk space, slows down Shippable and your local SBT startup times.


jarlah:
LGTM

PR #133, opened 2016-09-14
WIP: MUSARK-219

kpmeen:
There are still some work to be done to bring the service up to the same level as service_event and service_storage_admin. And there are some loose ends that I need to take care of before I'm ready to merge. But it's good to have the PR here so others can see what has been done (and not).

Key points that will be improved shortly:
- ~~The EventTypeRegistry will be made more specific to specific categories of events.~~
- More tests should be continuously added.
- Integration tests need to be (re)implemented/reviewed.
- ~~Need to extract test bootstrapping into a new library `music-test`~~
  - `common_test` should be removed.
- ~~Need to extract MusitResult ADT stuff into a new library `music-result`.~~
- Move operations need to be exposed as services.
- ~~EnvRequirement events should be generated upon storage node update.~~


kpmeen:
The PR is very close to being ready for merge now. Expect it hitting master either by the end of today, or early tomorrow morning.

PR #142, opened 2016-09-25
Clean-up 1

kpmeen:
First PR cleaning up a bit on the previous monster PR that introduced service_storagefacility. Mostly related to tests and SBT build scripts. Which also updates Play to 2.5.8.


jarlah (approved):
Ser ikke noe uten videre. LGTM

PR #155, opened 2016-09-29
Fix thing aggregate

jarlah:


kpmeen (approved):
This is urgent and will be merged before required checks are completed

PR #156, opened 2016-09-29
Postgres docker

jarlah:


kpmeen:
Merged...all _necessary_ tests succeeded. Only service_time was failing, which isn't relevant anymore.

PR #165, opened 2016-10-03
Musark 413

sveigl:


kpmeen:
This PR is closed because it has been included in a branch that collects all museumId changes.
See #178

PR #172, opened 2016-10-07
Musark 417

ellenjo:
With MuseumId in all endpoints at storage_facility


kpmeen:
This PR is closed because it has been included in a branch that collects all museumId changes.
See #178

PR #179, opened 2016-10-13
Musark 402 - Object Search

jarlestabell:


service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+class ObjectSearchIntegrationSpec @Inject() ( //    val dao: ObjectSearchDao
+) extends MusitSpecWithServerPerSuite {
+
+  implicit override val patienceConfig: PatienceConfig = PatienceConfig(
+    timeout = Span(15, Seconds),
+    interval = Span(50, Millis)
+  )
+
+  def doSearch(museumId: Int, museumNo: String, subNo: String, term: String, page: Int = 1, limit: Int = 25) = {
+    def encode(str: String) = UriEncoding.encodePathSegment(str, "utf-8")
```
kpmeen: This entire function can be written using the functions available on `wsUrl`. Which will handle all sorts of URL-encoding etc for you. So you can focus on writing the tests. Also note the formatting of very long argument lists in the below example. It's much easier to read when you have limited screen width available (as in github reviews).

``` scala
  def doSearch(
    museumId: Int,
    museumNo: String,
    subNo: String,
    term: String,
    page: Int = 1,
    limit: Int = 25
  ): WSResponse = {
    val url = s"""/museum/$museumId/objects/search"""
    wsUrl(url).withQueryString(
      "museumNo" -> museumNo,
      "subNo" -> subNo,
      "term" -> term,
      "page" -> s"$page",
      "limit" -> s"$limit"
    ).get().futureValue
  }
```

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  "ObjectSearch" must {
+    "find an object which exists, via museumNo" in {
+
+      val res = doSearch(1, "C666", "", "")
+      res.status mustBe 200
+      res.body must include("C666")
+      res.body must include("Øks")
+    }
```
kpmeen, line 42: Some more test cases verifying the behaviour of the different query parameters and response statuses that one can expect would be nice.

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
+/**
+ * Created by jarle on 05.10.16.
+ */
+import com.google.inject.Inject
+import dao.ObjectSearchDao
+import models.{MuseumIdentifier, MusitThing, ObjectAggregation, ObjectId}
```
kpmeen: Remove unused imports.

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
+/**
```
kpmeen: Please remove these auto generated comments from IntelliJ. All of the same information is available in the git history.
To disable these header comments being inserted into new files:
Go to "Preferences" -> "Editor" -> "File and Code Templates" and select the "Includes" tab.
Select the list entry called "File Header". Select all the text in the window to the right, and press the "OK" button.

And from now on you'll get nice and clean files :-)

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import dao.ObjectSearchDao
+import models.{MuseumIdentifier, MusitThing, ObjectAggregation, ObjectId}
+import no.uio.musit.test.MusitSpecWithServerPerSuite
+import org.scalatest.time.{Millis, Seconds, Span}
+import play.api.libs.json.Json
+import play.utils.UriEncoding
+
+import scala.language.postfixOps
+
+class ObjectSearchIntegrationSpec @Inject() ( //    val dao: ObjectSearchDao
```
kpmeen: remove trailing commented out code

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+ */
+import com.google.inject.Inject
+import dao.ObjectSearchDao
+import models.{MuseumIdentifier, MusitThing, ObjectAggregation, ObjectId}
+import no.uio.musit.test.MusitSpecWithServerPerSuite
+import org.scalatest.time.{Millis, Seconds, Span}
+import play.api.libs.json.Json
+import play.utils.UriEncoding
+
+import scala.language.postfixOps
```
kpmeen: not necessary since this feature is enabled in our build scripts (in `project/CommonSettings.scala`).

service_thing_aggregate/src/main/resources/evolutions/default/1.sql
```diff
@@ -23,3 +23,13 @@ CREATE TABLE MUSARK_STORAGE.STORAGE_NODE (
  museum_id INTEGER NOT NULL,
  PRIMARY KEY (storage_node_id)
 );
+
+CREATE TABLE MUSIT_MAPPING.MUSITTHING (
```
kpmeen, line 27: We also need this table in our `docker/dev/database/init.sql` script, for postgres.

kpmeen: Also, I get a bit confused sometimes. For example, this table is supposed to contain museum objects. Yet the table is called thing. Isn't thing an abstraction above object? Will there be anything else than objects here?
No need to answer in this thread though, I just needed to write down the question somewhere :-p

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.libs.json.{JsValue, Json}
+import play.api.mvc.{Action, Controller, Result, Results}
+import services.ObjectSearchService
+
+/**
+ * Created by jarle on 05.10.16.
+ */
+
+object Utils {
```
kpmeen: Please remove this Utils object entirely. There is only 1 controller method in this file, and it does not help for code readability that the actual result is calculated outside of the controller method.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+class ObjectSearchController @Inject() (
+    service: ObjectSearchService
+) extends Controller {
+
+  private val maxLimit = 100
+
+  def search(mid: Int, museumNo: String = "", subNo: String = "", term: String = "", page: Int = 1, limit: Int = 25) =
```
kpmeen: Please format very long attribute lists with each argument on separate lines:

``` scala
def foo(
    arg1: String,
    arg2: String,
    ...
    argN: String
) = ...
```

Also, since the attributes have default values defined in the routes file, there is no need to define default values here.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+    Action.async { request =>
+
+      val limitToUse = Math.max(limit, maxLimit)
```
kpmeen: This will always return `maxLimit`, unless `limit` is a bigger number. So if you pass in a limit of `25` (which is default), you will end up with up to `100` results. Something an integration test hopefully would have picked up.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+
+      def toJson(ob: Seq[MusitThing]) = Json.toJson(ob)
+
+      service.search(mid, museumNo, subNo, term, page, limitToUse).map(mr => Utils.musitResultToPlayResult(mr, toJson))
+    }
```
kpmeen, line 48: The entire method should be re-written to be something along the lines of:

``` scala
class ObjectSearchController @Inject() (service: ObjectSearchService) extends Controller {

  val logger = Logger(classOf[ObjectSearchController])

  private val maxLimit = 100
  private val defaultLimit = 25

  def search(
    mid: Int,
    museumNo: String,
    subNo: String,
    term: String,
    page: Int = 1,
    limit: Int = defaultLimit
  ) = Action.async { implicit request =>

    val limitToUse = limit match {
      case lim: Int if lim > maxLimit => maxLimit
      case lim: Int if lim < 0 => defaultLimit
      case lim: Int => lim
    }

    service.search(mid, museumNo, subNo, term, page, limitToUse).map {
      case MusitSuccess(res) =>
        Ok(Json.toJson[Seq[MusitThing]](res))

      case MusitDbError(msg, ex) =>
        logger.error(msg, ex.orNull)
        InternalServerError(Json.obj("message" -> msg))

      case err: MusitError =>
        logger.error(err.message)
        Results.InternalServerError(Json.obj("message" -> err.message))
    }
  }
}
```

service_thing_aggregate/src/main/scala/services/ObjectSearchService.scala
```diff
@@ -0,0 +1,19 @@
+package services
+
+/**
```
kpmeen, line 3: Remove auto generated comment

service_thing_aggregate/src/main/scala/services/ObjectSearchService.scala
```diff
@@ -0,0 +1,19 @@
...
+import dao.ObjectSearchDao
+import models.MusitThing
+import no.uio.musit.service.MusitResults.MusitResult
+
+import scala.concurrent.Future
+
+class ObjectSearchService @Inject() (
+    objectSearchDao: ObjectSearchDao
+) {
+  def search(mid: Int, museumNo: String = "", subNo: String = "", term: String = "", page: Int = 1, limit: Int = 25): Future[MusitResult[Seq[MusitThing]]] = {
```
kpmeen, line 16: formatting of very long argument list

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+import no.uio.musit.test.{MusitSpecWithAppPerSuite}
+import org.scalatest.time.{Millis, Seconds, Span}
+
+import no.uio.musit.service.MusitResults.MusitSuccess
+import models.{MuseumIdentifier, ObjectId}
+import no.uio.musit.test.MusitSpecWithAppPerSuite
+import org.scalatest.time.{Millis, Seconds, Span}
+import play.api.Logger
+
+/**
```
kpmeen, line 14: remove auto generated comment

service_thing_aggregate/src/test/scala/models/MuseumNoSpec.scala
```diff
@@ -0,0 +1,42 @@
+package models
+
+/**
+ * Created by jarle on 10.10.16.
```
kpmeen, line 4: Remove auto generated comment

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
+package models.dto
+
+import models.MusitThing
+
+/**
+ * Created by jarle on 06.10.16.
```
kpmeen, line 6: Remove auto generated comment

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+import models.MusitThing
+import no.uio.musit.service.MusitResults.{MusitDbError, MusitError, MusitResult, MusitSuccess}
+import play.api.Logger
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.libs.json.{JsValue, Json}
+import play.api.mvc.{Action, Controller, Result, Results}
+import services.ObjectSearchService
+
+/**
+ * Created by jarle on 05.10.16.
```
kpmeen: Remove auto generated comment

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
```diff
@@ -0,0 +1,49 @@
...
+object Utils {
+
+  //TODO: Move this to somewhere common
+  def musitResultToPlayResult[A](musitResult: MusitResult[A], toJsonTransformer: A => JsValue): Result = {
+    musitResult match {
+      case MusitSuccess(res) =>
+        Results.Ok(toJsonTransformer(res))
+
+      case MusitDbError(msg, ex) =>
+        Logger.error(msg, ex.orNull)
```
kpmeen: You should not use the logger in this way. The log message will be logged under "Application", and not the proper "package.name.ClassName" context. Making it much harder to make sense of the logs. See example below for proper usage.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+sealed trait FieldValue
+
+case class EmptyValue() extends FieldValue
+
+case class LiteralValue(v: String) extends FieldValue
+
+/*If v contains a value which needs to be escaped, escapeChar contains the appropriate escape character.
+If v doesn't contains a value which needs to be escaped, escapeChar =
```
kpmeen: Please use proper scala/javadoc formatting for these types of comments:

``` scala
/**
 * If v contains a value which needs to be escaped, escapeChar contains the
 * appropriate escape character.
 *
 * If v doesn't contains a value which needs to be escaped, escapeChar =
 */
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+ */
+case class WildcardValue(v: String, escapeChar: Char) extends FieldValue
+
+//#If we (in the future) need to treat some characters as invalid: case class InvalidValue(errorMessage: String) extends FieldValue
```
kpmeen: Remove

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    val dbConfigProvider: DatabaseConfigProvider
+) extends HasDatabaseConfigProvider[JdbcProfile] {
+
+  val logger = Logger(classOf[ObjectAggregationDao])
+
+  import driver.api._
+
+  private val musitThingTable = TableQuery[MusitThingTable]
+
+  def testInsert(museumId: Int, mthing: MusitThing): Future[Long] = {
```
kpmeen, line 36: This must be removed. Instead, if you need data in the tables for testing. Create insert statements for the data you need in the relevant `src/{it | test}/resources/evolutions/default/*.sql` file.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    val dto = MusitThingDto.fromDomain(museumId, mthing)
+
+    val action = musitThingTable returning musitThingTable.map(_.id) += dto
+    db.run(action)
+  }
+
+  val noEscapeChar = '\u0000' //Needs to be the same as Slicks no-escape-char value! (Default second parameter value to the like function)
```
kpmeen: Please keep long comments like these above the code being commented. Code formatters tend to automatically put long comments like these below the code its commenting. Which makes it hard to make sense of it.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  val escapeChar = '¤' //can be any char really. Public just to let the unit tests get at it
```
kpmeen, line 46: Unit test doesn't need to access this on the Dao at all. It is actually good if the test has a local definition of this property. If it changes here, we want the tests to fail. Because that means we will have data in the database using a different escape character than our code.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  def classifyValue(rawValue: String): FieldValue = {
+    //Since we build up a Slick Query object, we don't need to verify that the rawValue is "safe",
+    // the database engine will validate the parameter value.
+
```
kpmeen: Both of these blocks of comments should be placed above the function in a properly formatted comment block

```
  /**
   * Since we build up a Slick Query object, we don't need to verify that the
   * rawValue is "safe", the database engine will validate the parameter value.
   * So security-wise, we don't need to guard against '--' etc here. (Else we 
   * could have flagged values containing "--" etc as InvalidValue here)
   * 
   * We use '*' as wildcard symbol. And treat both '%' and '_' as ordinary
   * characters, both in like-tests and =-tests.
   */
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    if (rawValue.isEmpty) EmptyValue()
+    else if (rawValue.contains('*')) {
+      val wValue = rawValue
+        .replace(escapeChar.toString, s"${escapeChar}${escapeChar}")
+        .replace("%", s"${escapeChar}%")
+        .replace("_", s"${escapeChar}_")
+        .replace('*', '%')
+
+      /*Note that in the above expression, order is vital!
+        It is essential that the escapeChar -> escapeChar+escapeChar is done before the replacements which
```
kpmeen: It's common practice in both Scala and Java (and other languages I know), to comment _above_ the code being commented.
Also, the formatting should be:

```
      // Note that in the above expression, order is vital!
      // It is essential that the escapeChar -> escapeChar+escapeChar is done
      // before the replacements which introduces any escapeChars and that
      // %->escapeChar happens before *->'%'
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+
+      val esc = if (wValue.contains(escapeChar)) escapeChar else noEscapeChar
+
+      WildcardValue(wValue, esc)
+    } else {
+      LiteralValue(rawValue)
+    }
+  }
+
+  def fieldValueToMusitResult(fieldValue: FieldValue): MusitResult[FieldValue] = {
```
kpmeen: This function is redundant

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    MusitSuccess(fieldValue)
+  }
+
+  def classifyValueAsMusitResult(rawValue: String): MusitResult[FieldValue] = {
```
kpmeen, line 83: This function is redundant

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+        val sortedQuery = query.sortBy(mt => (mt.museumNoAsNumber.asc, mt.museumNo.asc /*Should we use toLowerCase on sorting on museumNo?*/ ,
+          mt.subNoAsNumber.asc, mt.subNo.asc, mt.id.asc)) //Should we search on term as well?
+
+        val offset = (page - 1) * pageSize
+        sortedQuery.drop(offset).take(pageSize)
+    }
+  }
+
+  //TODO: Remove this and use common similar implementation (when that is finished)
+  private def musitResultFutureToFutureMusitResult[A](value: MusitResult[Future[A]]): Future[MusitResult[A]] = {
```
kpmeen, line 155: This code is redundant, please remove

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    value match {
+      case MusitSuccess(succ) => succ.map(a => MusitSuccess(a))
+      case err: MusitError => Future.successful(err)
+    }
+  }
+
+  def search(mid: Int, museumNo: String, subNo: String, term: String, page: Int, pageSize: Int): Future[MusitResult[Seq[MusitThing]]] = {
```
kpmeen, line 162: This entire method can be written as follows:

``` scala
  def search(
    mid: Int,
    museumNo: String,
    subNo: String,
    term: String,
    page: Int,
    pageSize: Int
  ): Future[MusitResult[Seq[MusitThing]]] = {
    val query = searchQuery(mid, museumNo, subNo, term, page, pageSize).result

    db.run(query).map { seq =>
      MusitSuccess(seq.map(MusitThingDto.toDomain))
    }.recover {
      case e: Exception =>
        val msg = s"Error while retrieving search result"
        logger.error(msg, e)
        MusitDbError(msg, Some(e))
    }
  }
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+        case e: Exception =>
+          val msg = s"Error while retrieving search result"
+          logger.error(msg, e)
+          MusitDbError(msg, Some(e))
+      }
+    }
+    musitResultFutureToFutureMusitResult(resultFutSeq).map(_.flatten)
+  }
+
+  /** Gets the underlying SQL-statement that will be used to execute this search. This method is only meant for testing code! */
```
kpmeen, line 179: This must be removed. Testing this function in the unit test, doesn't test the actual search method. calling this doesn't really test anything other than the `searchQuery` method. Which can be tested by itself.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+      val action = query.result
+      action.statements.head
+    }
+  }
+
+  class MusitThingTable(
+      val tag: Tag
+  ) extends Table[MusitThingDto](tag, Some("MUSIT_MAPPING"), "MUSITTHING") {
+
+    def * = (museumId, id.?, museumNo, museumNoAsNumber, subNo, subNoAsNumber, term) <> (create.tupled, destroy) // scalastyle:ignore
```
kpmeen, line 191: Not overly important in this case, but it does help with readability on narrow viewports

``` scala
    // scalastyle:off method.name
    def * = (
      museumId,
      id.?,
      museumNo,
      museumNoAsNumber,
      subNo,
      subNoAsNumber,
      term
      ) <> (create.tupled, destroy)
    // scalastyle:on method.name
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+
+    val id = column[Long]("ID", O.PrimaryKey, O.AutoInc)
+    val museumId = column[Int]("MUSEUMID")
+    val museumNo = column[String]("MUSEUMNO")
+    val subNo = column[Option[String]]("SUBNO")
+    val term = column[String]("TERM")
+    val museumNoAsNumber = column[Option[Long]]("MUSEUMNOASNUMBER")
+    val subNoAsNumber = column[Option[Long]]("SUBNOASNUMBER")
+
+    def create = (museumId: Int, id: Option[Long], museumNo: String, museumNoAsNumber: Option[Long],
```
kpmeen, line 201: formatting on both create and destroy

``` scala
    def create = (
      museumId: Int,
      id: Option[Long],
      museumNo: String,
      museumNoAsNumber: Option[Long],
      subNo: Option[String],
      subNoAsNumber: Option[Long],
      term: String
    ) =>
      MusitThingDto(
        museumId = museumId,
        id = id,
        museumNo = museumNo,
        museumNoAsNumber = museumNoAsNumber,
        subNo = subNo,
        subNoAsNumber = subNoAsNumber,
        term = term
      )

    def destroy(thing: MusitThingDto) =
      Some((
        thing.museumId,
        thing.id,
        thing.museumNo,
        thing.museumNoAsNumber,
        thing.subNo,
        thing.subNoAsNumber,
        thing.term
      ))
```

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
...
+object MusitThingDto {
+
+  implicit def toDomain(x: MusitThingDto): MusitThing =
+    MusitThing(
+      museumNo = x.museumNo,
+      subNo = x.subNo,
+      term = x.term
+    )
+
+  val regExp = """\A\D*(\d+)(?:\D.*)?\z""".r // A sequence of non-digits, followed by a sequence of digits,
```
kpmeen, line 19: Comments should be placed above the line they are describing.

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
...
+  // followed by an optional tail starting with a non-digit (and then whatever).
+
+  /** The number part of a museumNo */
```
kpmeen, line 22:
```
/**
 * The number part of a museumNo
 */
```

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
+package models.dto
+
+import models.MusitThing
+
+/**
+ * Created by jarle on 06.10.16.
+ */
+case class MusitThingDto(museumId: Int, id: Option[Long], museumNo: String, museumNoAsNumber: Option[Long], subNo: Option[String], subNoAsNumber: Option[Long], term: String)
```
kpmeen, line 8: Format long args  with 1 arg per line

``` scala
case class MusitThingDto(
    museumId: Int,
    id: Option[Long],
    museumNo: String,
    museumNoAsNumber: Option[Long],
    subNo: Option[String],
    subNoAsNumber: Option[Long],
    term: String
)
```

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  val regExp = """\A\D*(\d+)(?:\D.*)?\z""".r // A sequence of non-digits, followed by a sequence of digits,
+  // followed by an optional tail starting with a non-digit (and then whatever).
+
+  /** The number part of a museumNo */
+  def museumNoNumberPart(museumNo: String): Option[Long] = {
+    val optM = regExp.findFirstMatchIn(museumNo)
+
+    optM.map { m =>
+      assert(m.groupCount == 1) //This regular expression is designed to only return one group
```
kpmeen, line 27: First of all...the two comments below should be above both of the statements. They way they are formatted now, as 3 lines to the right of (and below) the code, makes it _very_ difficult to read.

Secondly, `assert` is an `elidable` statement in Scala. Meaning, the compiler may decide to remove  the statement entirely if configured to do so.

IF this function should indeed throw an exception if the group count is > 1, it is much better to be explicit about it.

But, since this function is _only_ used in a test, this Dto, and through the `fromDomain` function below in the Dao. In a method that must be removed from there. You should remove this function as it is unnecessary for the purposes of object search.

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
...
+      m.group(1).toLong //Per def of this re, this should always be possible (within reasonable length of museumNo!)
+      // and never throw any exceptions.
+    }
+  }
+
+  def subNoNumberPart(subNo: String): Option[Long] = {
```
kpmeen, line 33: Not being used in production code outside the `fromDto`. Please remove.

service_thing_aggregate/src/main/scala/models/dto/MusitThingDto.scala
```diff
@@ -0,0 +1,48 @@
...
+    museumNoNumberPart(subNo)
+  }
+
+  def fromDomain(museumId: Int, x: MusitThing): MusitThingDto =
```
kpmeen, line 37: Not being used in production code outside the `fromDto`. Please remove.

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+
+  implicit override val patienceConfig: PatienceConfig = PatienceConfig(
+    timeout = Span(15, Seconds),
+    interval = Span(50, Millis)
+  )
+
+  val escapeChar = dao.escapeChar
+
+  def insertTestData(museumId: Int) = {
+    def insert(museumNo: String, subNo: String, term: String) = {
```
kpmeen, line 28: This entire block of code is using code in the Dao implementation that must be removed. Instead, to insert test data to test the search, add insert statements to the relevant `*.sql` scritps in the it/test evolutions folders.

``` sql
insert into MUSIT_MAPPING.MUSITTHING(museumNo, subNo, term, museumId, museumNoAsNumber, subNoAsNumber) values("C1", "1a", "Øks", 1, 1, 1);
insert into MUSIT_MAPPING.MUSITTHING(museumNo, subNo, term, museumId, museumNoAsNumber, subNoAsNumber) values("C1", "2a", "Skummel øks", 1, 1, 2);
insert into MUSIT_MAPPING.MUSITTHING(museumNo, subNo, term, museumId, museumNoAsNumber, subNoAsNumber) values("C1", "3", "Fin øks", 1, 1, 3);
```

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+
+  "ObjectSearch" must {
+    "dummy test to insert test data" in {
+
+    }
+    "find an object which exists, via museumNo" in {
+      insertTestData(1)
+
+      val res = dao.search(1, "C1", "", "", 1, 100).futureValue
+      val seq = res.get
```
kpmeen, line 95: You should also check that the result is what you expect it to be. Otherwise the test will fail with an ugly exception (because of trying to get a non existing element) instead of a proper description of what went wrong.

``` scala
val res = dao.search(1, "C1", "", "", 1, 100).futureValue
res.isSuccess mustBe true
res.get.length must be >= 3
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+import no.uio.musit.service.MusitResults._
+import play.api.Logger
+import play.api.db.slick.{DatabaseConfigProvider, HasDatabaseConfigProvider}
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import slick.driver.JdbcProfile
+
+import scala.concurrent.Future
+
+/**
+ * Created by jarle on 06.10.16.
```
kpmeen: Remove autogenerated comment

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+        val q1 = maybeAddMuseumNoFilter(musitThingTable, musemNoValue)
+
+        val q2 = maybeAddFilter[QMusitThingTable](q1, subNoValue, (q, value) => q.filter(_.subNo === value),
+          (q, value, esc) => q.filter(_.subNo like (value, esc)))
+
+        val q3 = maybeAddFilter[QMusitThingTable](q2, termValue, (q, value) => q.filter(_.term.toLowerCase === value.toLowerCase),
+          (q, value, esc) => q.filter(_.term.toLowerCase.like(value.toLowerCase, esc)))
+
+        val query = q3
+        val sortedQuery = query.sortBy(mt => (mt.museumNoAsNumber.asc, mt.museumNo.asc /*Should we use toLowerCase on sorting on museumNo?*/ ,
```
kpmeen, line 146: Remove this and below comment and pull arguments up to the same line, or use `{ }` instead of `( )` for the partial function on `sortBy`.

Please check with Line about how museumNo should be treated when sorting. And if the term should be searchable.

``` scala
query.sortBy { mt =>
  (mt.museumNoAsNumber.asc, mt.museumNo.asc, mt.subNoAsNumber.asc, mt.subNo.asc, mt.id.asc)
}
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+    MusitSuccess(fieldValue)
+  }
+
+  def classifyValueAsMusitResult(rawValue: String): MusitResult[FieldValue] = {
+    fieldValueToMusitResult(classifyValue(rawValue))
+  }
+
+  type QMusitThingTable = Query[MusitThingTable, MusitThingTable#TableElementType, scala.Seq]
+
+  private def maybeAddFilter[Q <: QMusitThingTable](q: Q, value: FieldValue, filterEqual: (Q, String) => Q, filterLike: (Q, String, Char) => Q): Q = {
```
kpmeen: Formatting of argument list. Hard to read code in review.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+        if (isAllDigits(v)) {
+          q.filter(_.museumNoAsNumber === v.toLong)
+        } else {
+          q.filter(_.museumNo.toUpperCase === v.toUpperCase)
+        }
+      case WildcardValue(v, esc) => q.filter(_.museumNo.toUpperCase like (v.toUpperCase, esc))
+    }
+  }
+
+  def searchQuery(mid: Int, museumNo: String, subNo: String, term: String, page: Int, pageSize: Int): MusitResult[QMusitThingTable] = {
```
kpmeen: Place each argument on a separate line as commented previously.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+        subNoValue <- classifyValueAsMusitResult(subNo)
+        termValue <- classifyValueAsMusitResult(term)
+      } yield (museumNoValue, subNoValue, termValue)
+
+    triple.map {
+      case (musemNoValue, subNoValue, termValue) =>
+
+        val q1 = maybeAddMuseumNoFilter(musitThingTable, musemNoValue)
+
+        val q2 = maybeAddFilter[QMusitThingTable](q1, subNoValue, (q, value) => q.filter(_.subNo === value),
```
kpmeen: This is very hard to read because the argument list is broken to a new line in the middle. It's not intuitively readable as the same argument list because of all the `(` and `)` characters.

More readable formatting (with named arguments) is:

``` scala
        val q2 = maybeAddFilter[QMusitThingTable](
          q = q1,
          value = subNoValue,
          filterEqual = (q, value) => q.filter(_.subNo === value),
          filterLike = (q, value, esc) => q.filter(_.subNo like (value, esc))
        )
```

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
```diff
@@ -0,0 +1,204 @@
...
+          (q, value, esc) => q.filter(_.subNo like (value, esc)))
+
+        val q3 = maybeAddFilter[QMusitThingTable](q2, termValue, (q, value) => q.filter(_.term.toLowerCase === value.toLowerCase),
```
kpmeen: Same as above

service_thing_aggregate/src/main/scala/models/MusitThing.scala
```diff
@@ -0,0 +1,13 @@
+package models
+
+import play.api.libs.json.Json
+
+/**
+ * Created by jarle on 06.10.16.
```
kpmeen, line 6: Remove auto-generated comment

service_thing_aggregate/src/main/scala/models/MusitThing.scala
```diff
@@ -0,0 +1,13 @@
...
+ */
+case class MusitThing(museumNo: String, subNo: Option[String], term: String)
```
kpmeen, line 8: I guess I have the same question here as previously with regards to the naming. Is this supposed to be a `MusitThing` or a `MusitObject`?

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq.length mustBe 1 //We should find C81¤A and *not* C81¤XA
+    seq(0).museumNo mustBe s"C81${escapeChar}A"
+  }
+
+  "that the esacpe character is treated like an ordinary character in like-context" in {
+    val res = dao.search(1, s"C*${escapeChar}A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe s"C81${escapeChar}A"
```
kpmeen, line 227: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81-A"
+  }
+
+  "that the escape character is treated like an ordinary character in =-context" in {
+    val res = dao.search(1, s"C81${escapeChar}A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81¤A and *not* C81¤XA
+    seq(0).museumNo mustBe s"C81${escapeChar}A"
```
kpmeen, line 219: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81-A"
+  }
+
+  "that - is treated like an ordinary character in like-context" in {
+    val res = dao.search(1, "C*-A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81-A"
```
kpmeen, line 211: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81%A"
+  }
+
+  "that - is treated like an ordinary character in =-context" in {
+    val res = dao.search(1, "C81-A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81-A"
```
kpmeen, line 203: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81%A"
+  }
+
+  "that % is treated like an ordinary character in like-context" in {
+    val res = dao.search(1, "C*%A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81%A"
```
kpmeen, line 195: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq(0).museumNo mustBe "C888_A"
+    seq(1).museumNo mustBe "C888_B"
+  }
+
+  "that % is treated like an ordinary character in =-context" in {
+    val res = dao.search(1, "C81%A", "", "", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 1 //We should find C81%A and *not* C81%XA
+    seq(0).museumNo mustBe "C81%A"
```
kpmeen, line 187: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  "museumNo with * and _ should work" in {
+
+    val res = dao.search(1, "c888_*", "", "øks", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 2
+
+    seq(0).museumNo mustBe "C888_A"
```
kpmeen, line 178: `seq.head`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+    seq(1).museumNo mustBe "C888_B"
```
kpmeen, line 179: `seq.last`

service_thing_aggregate/src/test/scala/dao/ObjectSearchDaoSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+
+  "museumNo, subNo with * and term search should work" in {
+
+    val res = dao.search(1, "c555*", "3*", "øks", 1, 10).futureValue
+    val seq = res.get
+
+    seq.length mustBe 3
+    seq(0).subNo mustBe Some("34A")
+    seq(1).subNo mustBe Some("34B")
+    seq(2).subNo mustBe Some("34C")
```
kpmeen, line 168: FYI, this can be written in a more concise way...although readability is probably not as good.

``` scala
seq.map(_.subNo) must contain allOf(Some("34A"), Some("34B"), Some("34C"))
```

kpmeen (changes requested):
A few of the comments are minor issues regarding autogenerated comments. These are of lowest priority to take care of. I've added them mostly to raise awareness. It would be great if they were removed though :-)

Apart from that, there are quite a few problems with formatting, unnecessary code, test specific code in production code and a bug related to the `limit` parameter in the controller.

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Ok. (I assume you meant delete the text in the File Header text window.)

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Ok

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Have removed postfixOps.
(No idea where it came from, probable copy+paste from somewhere else)

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Ok

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
kpmeen: yes, that was it :-)

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Ok

service_thing_aggregate/src/it/scala/ObjectSearchIntegrationSpec.scala
jarlestabell: Ok, added a couple more. (But only 200 is returned by the endpoint in non-internal-error circumstances, so only 200-tests are done)

service_thing_aggregate/src/main/resources/evolutions/default/1.sql
jarlestabell: I've added the table to postgres. (Hopefully I'll remember to ask Svein or Rituvesh to help me start the stuff, to check that I didn't mess up the init.sql file!)
Natural history people use the term 'Object'. The Cultural side don't use that term and prefer 'Gjenstand'. 'Thing' is sort of a compromise. It is also what CIDOC CRM has ended up using.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
jarlestabell: Ok

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
jarlestabell: I had it separately because I believe that something like this ought to be in some common file. To go from MusitResult to (Play)Result is something we have to do in all end points. It's too easy to screw this up, so we should have a standard way of handling it.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
jarlestabell: Aha. I copied from ObjectAggregationController, so there may be a bug there as well.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
jarlestabell: True. :)

jarlestabell: But should we really manually enter more than 100 objects in the SQL-script?

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
jarlestabell: Ok, have done that.
But that error handling code is rather brittle and will probably need to change before going into production.
(I don't like to copy+paste code which will become buggy in the future when something changes further down in the layers. Especially since the compiler won't notify us about future problems here.)

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: I understand that it doesn't belong in the dao file. But could we instead create it in a separate TestObjectSearchDao file?
"Manual" SQL insert statements is generally more work (one has to do the business processing in our heads instead of in the code) and with multiple test files each contributing different set of SQL-inserts makes it rather hard to tell which tests depends upon which data, making the test code rather brittle.
Doing it programmatically also enables generating bigger test data (like more than 100 rows to test that limits work)

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: I think there's a misunderstanding. The tests wants to use this character deliberately, in order to actually test the escape character logic. Not using the proper escapeChar in the tests make those tests "impotent". (The tests don't fail if the test uses the wrong escape character, on the contrary, it makes the escape character tests "more likely" to succeeed).

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: True. It had meaning early on where certain characters were classified as illegal. (Like '%')

service_thing_aggregate/src/main/resources/evolutions/default/1.sql
lasjo: We have an agreement between cultural and natural history to use the term 'Object' (in some cases 'museum object') as a common term. 'Thing' was intended as a common term for objects, documents, photographs, and so forth.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: As mentioned above, early on we considered '%' illegal. Currently, the code considers no character illegal. If this is ok, then more of the code can be simplified.
If I understand it correctly, since we squash commits during PRs, earlier commits in a PR become lost?
That's why there's some redundant stuff there because I wasn't totally sure it would be ok/safe to allow all characters.
(If there was access to previous commits in the PR, I could have simplified the code to the core and instead linked to an earlier version in a comment, in case one wants to change the decisions about allowing everything)

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
jarlestabell: Ok.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
kpmeen: Well, the thing is that we cannot say that there will be a standard way of doing this. Because each controller method may decide for itself what to do in case of success or error. Assuming `Results.Ok`, or `Results.InternalServerError` will only ever cover a subset of use-cases. We have tests to ensure that the results are returned as expected.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
kpmeen: That has slipped my dwindling eyesight, unfortunately. But yes, that controller too should use the suggested construct.

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
kpmeen: No, that shouldn't be necessary. What you can do to make this easier to test, is to make the `maxLimit` property configurable through the `application.conf` file in `service_thing_aggregate`. That way you can set the property to a much lower value during testing to avoid loading 100s of rows of data.

In `application.conf` you can add the following property at the end of the file:

```
musit.thing.search.max-limit = 100
```

When setting up the spec (for e.g. the integration testing), you can add a config file in `src/it/resources/application.conf` with the same config parameter, but with a lower value. When starting the tests, this value will be used instead of the one defined in `src/main/resources/application.conf`. (Same applies in `src/test/resources/application.conf`)

service_thing_aggregate/src/main/scala/controllers/ObjectSearchController.scala
kpmeen: No it is _not_ brittle. It is specific to this controller method. And as I have mentioned previously, it is not bad practice to have some duplication of code. Especially when it is essential to understand the behaviour of the method where its used.

When the code is properly covered by tests, they will notify us about the change in behaviour.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
kpmeen: Short answer is; no we can't add test related code into the production code base.

Instead, you can as mentioned in a comment above, test with a limit that is much lower than 100 to verify the behaviour.

Eventually though, we should have a better way of adding data to our unit and integration tests . But for now, it is sufficient to use inserts in this case. Since the service is read only, its not wrong to assume that data is inserted using the evolution scripts in test and it. If you need to clarify where the data comes from, you can add a properly formatted comment to the test class.

service_thing_aggregate/src/main/scala/dao/ObjectSearchDao.scala
kpmeen: The commit is only "lost" on the master branch when merged to create a more linear history. Not on the branch it was created on. What typically happens is that the branch used is removed when no longer needed. So in that respect, the commit is "lost" on github. However, on your machine you will still have the same branch with full history.

This is one of many reasons why working on forks are consider to be a better approach. It allows you to keep the history _you_ want to keep. Isolated and separately from the upstream repository.

In production code however, only code and comments that are relevant to the execution of the system should be present. And commented out code is typically not merged in.

kpmeen:
Closing this PR. New will come where additional data is included in response (objectId, currentLocationId, path and pathNames).

PR #186, opened 2016-10-20
MUSARK-402

kpmeen:
Implements object search according to Jira issue MUSARK-400.
Results include `currentLocationId`, `path` and `pathNames`.


jarlah (approved):
Looks good

PR #187, opened 2016-10-20
Search results obviously needs to return the total number of matches …

kpmeen:
…to be able to calculate the paging properly


jarlah (approved):
lgtm

PR #189, opened 2016-10-21
MUSARK-200 and MUSARK-491

kpmeen:
Adds rules for validating the basic structure of a node hierarchy and whether or not a node can be moved or added to a given destination.


kpmeen:
Still need to create some negative tests to verify that the validation works as expected for move events. But all else should be well covered.

PR #201, opened 2016-11-09
MUSARK-66

kpmeen:
Adding restrictions to service endpoints that dictates which operations users are allowed to perform based on their auth level for a given museum.

### DIFF
--- a/musit-service/src/main/scala/no/uio/musit/security/Roles.scala
+++ b/musit-service/src/main/scala/no/uio/musit/security/Roles.scala
@@ -37,6 +37,7 @@ object Roles {
     permissions: Seq[Permission]
   )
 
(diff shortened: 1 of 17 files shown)


musit-service/src/main/scala/no/uio/musit/security/Roles.scala
```diff
@@ -46,12 +47,16 @@ object Roles {
   val UmSfRead = "UmSfRead"
   val UmSfWrite = "UmSfWrite"
   val UmSfAdmin = "UmSfAdmin"
+  val VmSfRead = "VmSfRead"
+  val VmSfWrite = "VmSfWrite"
+  val VmSfAdmin = "VmSfAdmin"
```
kpmeen, line 52: You're drunk, Codacy!

PR #202, opened 2016-11-09
MUSARK-552

kpmeen:
Refactoring Ids related to actors and users.


kpmeen:
It's important to note that _all_ user/actor ID's now are expected to be `UUID`s!

kpmeen:
I have not yet updated the init.sql script for postgres!

jarlah:
looks good to me.

PR #213, opened 2016-11-26
Oracle

jarlah:


kpmeen (commented):
We need to find a way to ensure the docker-compose script is using an authenticated dockerhub user with contribution rights to the private docker repo. Ideally the image should be owned by a `musit` user.

kpmeen (approved):

PR #220, opened 2016-12-05
Fix nginx

jarlah:


docker/utv/nginx/nginx.conf
```diff
@@ -1,57 +1,49 @@
...
     }
 
-    server {
-        listen 80;
-        server_name	w3utv-mu01.uio.no;
-        if ($request_uri != "/check.txt") {
-            rewrite ^(.*)$ 	https://musit-utv.uio.no$1 permanent;
-        }
+    location /api/auth/ {
+      proxy_pass http://auth:7070/;
```
kpmeen: Should only expose the `rest` endpoints until we have secured the `web` endpoints.

kpmeen (changes requested):

kpmeen (approved):

PR #228, opened 2016-12-12
Adding service for looking up storage node based on UUID.

kpmeen:


kpmeen:
Endpoint is now specified and service can be used for verifying the client implementation. Still need to implement some simple tests before merging.

PR #245, opened 2017-01-15
MUSARK-746: Move handling of login/out to the backend

kpmeen:
WIP: This is currently a work in progress. However there is not much left to do. From a client impl perspective the services are contractually complete, and can be used to begin the necessary refactoring. Also, the NodeJS `backing` module can then finally be removed.

*URI's*:
* login : `GET /api/auth/rest/authenticate`
* logout: `GET /api/auth/rest/logout`


kpmeen:
I've added a way to validate the Delphi clients in the old way, so we can have a gradual update/rollout of them...if necessary. The code specific to that is pretty much isolated into 1 method. And can easily be removed when there are no clients using the old way.

I'm _not_ happy about adding that piece of code!

On a different note. Seing that we will be getting a lot of sessions, it might be sensible to try and recycle `SessionUUID`s (where possible) to prevent rapid growth of data. Either that, or we enable a background process that can prune inactive sessions that are older than some predefined period of time.

jarlah (approved):

PR #253, opened 2017-02-09
Re-write of service for moving nodes. Correcting timestamp format in logback configs.

kpmeen:
Fixing the same implementation problem for moving nodes as MUSARK-819 did for moving objects.
Moving nodes is slightly more complex due to business rules validations etc. But at least now the two move operations use the same function to perform the actual move.

Also changed formatting of the logging timestamps on @halvoru's request. They should now comply with the ISO date standard. Don't rightly know why there was an `_` character in the format to start with.


kpmeen:
Verified and works as expected with manual tests locally as well.

PR #256, opened 2017-02-20
MUSARK-863: Preparation of `service_analysis` module

kpmeen:
Preparing project for development on new module for handling analysis of museum objects.


service_analysis/app/controllers/AnalysisController.scala
```diff
@@ -0,0 +1,129 @@
...
+
+@Singleton
+class AnalysisController @Inject() (
+    val authService: Authenticator,
+    val analysisService: AnalysisService
+) extends MusitController {
+
+  val logger = Logger(classOf[AnalysisController])
+
+  lazy val internalErr = (msg: String) => InternalServerError(Json.obj("message" -> msg))
```
ingarabr: We're only using this in context of `MusitError`, should we have changed the signature to a `MusitError`? It will make it less generic.

service_analysis/app/models/SampleObject.scala
```diff
@@ -0,0 +1,25 @@
...
+  createdDate: DateTime,
+  // TODO: Add autogenerated sample ID
+  sampleId: Option[String],
+  externalId: Option[String],
+  note: Option[String],
+  updatedBy: Option[ActorId],
+  updatedDate: Option[DateTime]
+)
+
+object SampleObject {
```
ingarabr: Unused object

service_analysis/app/repositories/dao/AnalysisDao.scala
```diff
@@ -0,0 +1,206 @@
...
+   * @param a The Analysis to persist.
+   * @return eventually returns a MusitResult containing the EventId.
+   */
+  def insert(a: Analysis): Future[MusitResult[EventId]] = {
+    val action = insertAnalysisWithResultAction(a)
+
+    db.run(action.transactionally).map(MusitSuccess.apply).recover {
+      case NonFatal(ex) =>
+        val msg = s"An unexpected error occurred inserting an analysis event"
+        logger.error(msg, ex)
```
ingarabr: Have we defined any guidelines for logging? Imo this kind of faults should be logged as warning.

service_analysis/app/repositories/dao/SampleObjectDao.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  def update(so: SampleObject): Future[MusitResult[Int]] = {
+    val a = sampleObjTable.filter(_.id === so.objectId).update(asSampleObjectTuple(so))
+
+    db.run(a.transactionally).map {
+      case res: Int if res == 1 => MusitSuccess(res)
+      case res: Int if 1 > res => MusitDbError("Nothing was updated")
+      case res: Int if 1 < res => MusitDbError(s"Too many rows were updated: $res")
```
ingarabr: In the case that we have to many rows updated, should we do a rollback? I find it strange that we're not doing it since we have done something and send an error to the caller.

service_analysis/app/controllers/AnalysisController.scala
kpmeen: Probably :-)

service_analysis/app/models/SampleObject.scala
kpmeen: In Progress

service_analysis/app/repositories/dao/AnalysisDao.scala
kpmeen: No, we don't, yet.

With respect to warning vs error...it depends. But we can have that discussion offline :-)

service_analysis/app/repositories/dao/SampleObjectDao.scala
kpmeen: It's a very hypothetical situation, really. Very much a legacy of earlier written code from before I came onboard. But, yes...if we're updating more than 1 row, we should probably do a rollback.
Let's talk about that tomorrow.

PR #263, opened 2017-03-09
MUSARK-925 Use MuseumId in dao

jarlah:
I have updated evolution scripts in  test to contain museumId for the two actors. And I am just passing down MuseumId which is already a required query parameter to the search endpoint.


jarlah:
Kommentar på  denne saken av line: 
NB må tenke på om det har noe å si at flere actors har samme UUID

er dette noe vi skal tenke på? hvordan kan det være et problem? og har det noe med å skille på museumId?

service_actor/app/repositories/dao/ActorDao.scala
```diff
@@ -47,10 +47,10 @@ class ActorDao @Inject() (
     db.run(query.result.headOption)
   }
 
-  def getByName(searchString: String): Future[Seq[Person]] = {
+  def getByName(mid: MuseumId, searchString: String): Future[Seq[Person]] = {
     val likeArg = searchString.toLowerCase
     val query = actorTable.filter { a =>
-      a.fn.toLowerCase like s"%$likeArg%"
+      (a.fn.toLowerCase like s"%$likeArg%") && a.museumId === mid.underlying
```
ingarabr: You can remove the `underlying` call by changing the setup of the table to read the Int to a `MuseumId` (line 82 in this file). You also need to add a Column type for `MuseumId` in `ColumnTypeMappers`

kpmeen: Mmm...I also see we have a job to do to update the return types of the public methods here. But that's an issue in itself, really.

service_actor/app/repositories/dao/UserInfoDao.scala
```diff
@@ -20,7 +20,7 @@
 package repositories.dao
 
 import com.google.inject.Inject
-import no.uio.musit.models.ActorId
+import no.uio.musit.models.{ActorId, MuseumId}
```
kpmeen: Unused import?

service_actor/test/resources/evolutions/test/1.sql
```diff
@@ -59,11 +59,11 @@ VALUES ('12345678-adb2-4b49-bce3-320ddfe6c90f', 'feide:fred@foo.com', 'Fred Flin
 INSERT INTO MUSARK_AUTH.USER_INFO (USER_UUID, SECONDARY_ID, NAME, EMAIL)
 VALUES ('d5a9a938-0905-470b-a892-88d0e7359579', 'feide:barney@foo.com', 'Barney Rubble', 'b.rubble@foo.bar.com');
 
-INSERT INTO MUSIT_MAPPING.ACTOR (ACTORNAME, DATAPORTEN_UUID, DATAPORTEN_USERNAME, APPLICATION_UUID)
-VALUES ('And, Arne1', null, 'andarn', '41ede78c-a6f6-4744-adad-02c25fb1c97c');
+INSERT INTO MUSIT_MAPPING.ACTOR (ACTORNAME, DATAPORTEN_UUID, DATAPORTEN_USERNAME, APPLICATION_UUID, MUSEUM_ID)
```
kpmeen, line 62: Have you double checked with @ellenjo that the `MUSEUM_ID` column is available in `MUSIT_MAPPING.ACTOR` table?

kpmeen (changes requested):

service_actor/test/resources/evolutions/test/1.sql
jarlah: ill check this tomorrow. If not then this is pretty useless work :)

kpmeen: Checked the `MUSIT_MAPPING.ACTOR` table just now, and there is indeed a column for museumId there. However, you'll need to change it's spelling.
But apparently it's not called `MUSEUM_ID` in that particular table, but `MUSEUMID`.

kpmeen: Checked the `MUSIT_MAPPING.ACTOR` table on UTV now, and the column is indeed there.

kpmeen (approved):

PR #267, opened 2017-03-16
reduce db connections to the database

ingarabr:
In our local dev environment with the current setup we ended up
with to many connection to the oracle database. Since we have
a lot of services where some of them doesn't have so much load
it's a bit overkill with the play/slick default settings. The default
settings for slick is:
execution threads: 20
min connection: 20
max connection 100 (20 *5)

This gives us a connection pool between 100 to 700 connection
for the 7 services. With the new settings in this commit we will
 end up with between 26 to 130 connections.


kpmeen (approved):

jarlah (approved):
LGTM

PR #269, opened 2017-03-16
Feature/reduce get calls in tests

ingarabr:
No more `get` calls in our test code. Codacy should now hava a better status :)


jarlah (approved):
LGTM

PR #271, opened 2017-03-17
MUSARK-885: Changing contract of moveObjects service to take id + type for objects to be moved

kpmeen:
This changes the contract of the web-client, while leaving the Delphi client untouched. The web-client will continue to work with the "old" API after this is merged, since it basically will treat all moved objects as collection objects. However, when the web-client needs to move sample objects as well, it will need to use the new contract for _all_ move objects requests.

example for new request contract is as follows:

```json
{
  "doneBy": "00000000-0000-0000-0000-000000000000",
  "destination": 5,
  "items": [{
    "id": 3,
    "objectType": "collection"
  },{
    "id": 5,
    "objectType": "collection"
  },{
    "id": 3,
    "objectType": "sample"
  }]
}
```

example for new response contract is:

```json
{
  "moved": [{
    "id": 3,
    "objectType": "collection"
  },{
    "id": 3,
    "objectType": "sample"
  }],
  "failed": [{
    "id": 5,
    "objectType": "collection"
  }]
}
```


kpmeen:
Note...this requires an updated oracle12c image, which is already created and pushed.

service_storagefacility/app/controllers/StorageController.scala
```diff
@@ -484,8 +484,8 @@ final class StorageController @Inject()(
   def search(
       mid: Int,
       searchStr: Option[String],
-      page: Int = 1,
```
jarlah, line 487: I cant see any route changes that can legitimate this change. This could break if the client (uncorrectly) uses the endpoint without limit or page. Optionally you could move the default values to routes?

service_storagefacility/app/models/event/move/MoveEvents.scala
```diff
@@ -73,13 +89,17 @@ case class MoveNode(
     eventType: EventType,
     from: Option[StorageNodeDatabaseId],
     to: StorageNodeDatabaseId
-) extends MoveEvent
+) extends MoveEvent {
+
+  override val objectType: ObjectType = Node
```
jarlah, line 94: Isn't this a code smell? That we need to relate to objectType when moving a node? Or is this intended? If so, should it be documented?

jarlah (changes requested):
Small things otherwise ok

service_storagefacility/app/models/event/move/MoveEvents.scala
jarlah: This might not be relevant comment

kpmeen: This is necessary to make the deeper levels of ugly Dao code work as expected. Not necessarily code smell since it isn't something any consumer of API's (external or internal) are exposed to.

service_storagefacility/app/controllers/StorageController.scala
kpmeen: I would double check your statement here. The line you comment was modified to have the default value explicitly stated in the routes file instead of both places (as it used to have) ;-)

jarlah: I didnt see any changes to the routes file for this code. So then i assume the routes file had default values then ? I did in fact not take the time to look at Either the branch or master to see if route file had defaults. So the review was purely based on the diff in the PR.

service_storagefacility/app/models/event/move/MoveEvents.scala
jarlah: Ok

service_storagefacility/app/controllers/StorageController.scala
kpmeen: That is correct :-)

jarlah (approved):

PR #273, opened 2017-03-20
MUSARK-882 Chore/database setup

ingarabr:


steinaol (approved):
LGTM. If I understand you correctly, there is no need for defining SLICK_DRIVER in deploy and docker-compose, because you instead include common definition files in the services?

ingarabr:
@steinaol ; that's right

PR #274, opened 2017-03-20
MUSARK-944 - Feature/configurable endpoint urls

ingarabr:
Make the application ready to be configured with a proxy for all external http calls.


musit-service/src/main/scala/no/uio/musit/security/dataporten/DataportenAuthenticator.scala
```diff
@@ -57,19 +71,12 @@ class DataportenAuthenticator @Inject()(
   private type DataportenToken = BearerToken
   private type AuthResponse    = Future[Either[Result, UserSession]]
 
-  private val defaultSessionTimeout = (2 hours).toMillis
```
kpmeen, line 60: The default session timeout of 2 hours is quite important. But I can't see that it's being taken care of in the changes you've made?

kpmeen (dismissed):

musit-service/src/main/scala/no/uio/musit/security/dataporten/DataportenAuthenticator.scala
ingarabr: It's covered by the `reference.conf` file.

musit-service/src/main/scala/no/uio/musit/ws/ProxiedRequest.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  val proxy: Option[WSProxyServer] = for {
+    host <- config.getString(ProxyHost)
+    port <- config.getInt(ProxyPort)
+    user     = config.getString(ProxyUser)
+    password = config.getString(ProxyPassword)
+  } yield
+    DefaultWSProxyServer(host = host, port = port, principal = user, password = password)
+
+  implicit def viaProxy(): WSRequest = proxy.map(req.withProxyServer).getOrElse(req)
```
kpmeen: Empty parens are typically used when you have side-effecting function calls like `doStuff()` etc. So if they aren't absolutely necessary the parens should be removed.

Also , I don't think this function needs to be implicit(?).

musit-service/src/main/scala/no/uio/musit/ws/ProxiedRequest.scala
```diff
@@ -0,0 +1,35 @@
...
+  val ProxyPassword = "ws.proxy.password"
+
+}
+
+trait ViaProxy {
+
+  implicit def viaProxy(req: WSRequest)(implicit config: Configuration): ProxiedRequest =
+    new ProxiedRequest(req, config)
+
+}
```
kpmeen, line 35: It's a good idea to also add an `object` of type `ViaProxy` to have a singleton available for direct import of the implicit converter.

That way you don't need to have strange extensions on the `class` definitions in cases where you need a `WSClient` with the `viaProxy` function available in implicit scope. Instead it will be implicitly available from the imported object.

See e.g. `no.uio.musit.formatters.WithDateTimeFormatters` and `no.uio.musit.formatters.DateTimeFormatters`

service_geo_location/app/services/GeoLocationService.scala
```diff
@@ -1,45 +1,34 @@
...
-import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.{Configuration, Logger}
 
 import scala.concurrent.Future
 
-class GeoLocationService @Inject()(config: Configuration, ws: WSClient) {
+case class GeoLocationConfig(url: String, hitsPerResult: Int)
+
+class GeoLocationService @Inject()(ws: WSClient)(implicit config: Configuration)
+    extends ViaProxy {
```
kpmeen: See above comment about creating an object that extends the  `ViaProxy` trait to allow the implicit scope to be imported instead of extending the class.

musit-service/src/main/scala/no/uio/musit/security/dataporten/DataportenAuthenticator.scala
```diff
@@ -46,30 +42,24 @@ import scala.concurrent.duration._
...
-    conf: Configuration,
     authResolver: AuthResolver,
-    ws: WSAPI
-) extends Authenticator
-    with OAuth2Constants {
+    ws: WSClient
+)(implicit conf: Configuration)
+    extends Authenticator
+    with OAuth2Constants
+    with ViaProxy {
```
kpmeen: See below comment about creating object instead of import the `ViaProxy` trait.

kpmeen (changes requested):
Just a couple of small things. Otherwise it looks good.

kpmeen (approved):

PR #277, opened 2017-03-22
MUSARK-957 and MUSARK-958

kpmeen:
Adding objectType to response for object location history service. Adding UUID to the response for the get objects service. Ensuring the objectType query param is used when filtering currentObjectLocation.


kpmeen:
NOTE: I need to publish a new docker image for oracle-12c for this to work locally! So do not pull down this change until the image is published.

service_storagefacility/app/services/StorageNodeService.scala
```diff
@@ -607,6 +608,8 @@ class StorageNodeService @Inject()(
             registeredDate = e.registeredDate.get,
             doneBy = e.doneBy,
             doneDate = e.doneDate,
+            id = e.affectedThing.get,
```
kpmeen, line 611: This get is OK since the `affectedThing` is a required field in the database in this context.

PR #279, opened 2017-03-23
MUSARK-945 - feature/label printer

ingarabr:
Tested with TSC and Brady printers with the following label sizes:

Brady: 52 x 27 mm
TSC: 54 x 26 mm


service_barcode/app/models/TemplateConfigs.scala
```diff
@@ -103,4 +105,12 @@ object TemplateConfigs {
     override val colsPerPage: Int         = 1
   }
 
+  case object LabelPrinter extends TemplateConfig {
+    override val templateId: Int          = 6
+    override val name: String             = "Label Printer"
```
ingarabr, line 110: Should we add the label size to the name?

jarlah (approved):

service_barcode/app/models/TemplateConfigs.scala
jarlah: not sure if it matters

kpmeen (approved):

PR #280, opened 2017-03-24
MUSARK-959 - feature/uniform urls

ingarabr:
Frontend PR https://github.com/MUSIT-Norway/musit-frontend/pull/353

### DIFF
--- a/docker/dev/nginx/nginx.conf
+++ b/docker/dev/nginx/nginx.conf
@@ -57,6 +57,10 @@ http {
       proxy_pass http://geolocation:7070/;
     }
 
+    location /api/geolocation/v1/ {
+      proxy_pass http://geolocation:7070/;
+    }
+
     location /api/thingaggregate/ {
       proxy_pass http://thingaggregate:7070/;
     }
--- a/docker/dev/nginx/nginx.conf
+++ b/docker/dev/nginx/nginx.conf
@@ -81,6 +81,10 @@ http {
       proxy_pass http://management:7070/service_management/;
     }
 
+    location /api/management/ {
+      proxy_pass http://management:7070/service_management/;
+    }
+
     # redirect server error pages to the static page /50x.html
     error_page   500 502 503 504  /50x.html;
     location = /50x.html {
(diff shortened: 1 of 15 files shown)


jarlah (commented):
Not really convinced that its the right solution to continue support v1 urls. The best is to demand changes in the client

kpmeen:
@jarlah we can't do that due to the Delphi client depending on the moveObject service. So until we know all installations of the KHM app has been updated with new URL, we have to have the v1 path in the config. The web client should not be allowed to use the v1 uri's though.

docker/dev/nginx/nginx.conf
```diff
@@ -54,19 +54,31 @@ http {
...
+      proxy_pass http://thingaggregate:7070/service_thing_aggregate/;
     }
 
     location /api/actor/ {
-      proxy_pass http://actor:7070/;
+      proxy_pass http://actor:7070/service_actor/;
+    }
+
+    location /api/actor/v1/ {
+      proxy_pass http://actor:7070/service_actor/;
```
kpmeen: Only the storagefacility service should have a v1 mapping enabled. The only reason is for the Delphi client not to break.

docker/dev/nginx/nginx.conf
```diff
@@ -54,19 +54,31 @@ http {
     }
 
     location /api/geolocation/ {
-      proxy_pass http://geolocation:7070/;
+      proxy_pass http://geolocation:7070/service_geo_location/;
+    }
+
+    location /api/geolocation/v1/ {
+      proxy_pass http://geolocation:7070/service_geo_location/;
```
kpmeen: See below comment

docker/utv/nginx/nginx.conf
```diff
@@ -63,19 +63,31 @@ http {
     }
 
     location /api/geolocation/ {
-      proxy_pass http://geolocation:7070/;
+      proxy_pass http://geolocation:7070/service_geo_location/;
+    }
+
+    location /api/geolocation/v1/ {
+      proxy_pass http://geolocation:7070/service_geo_location/;
```
kpmeen: See comment in dev nginx above

kpmeen (changes requested):

kpmeen:
@ingarabr please rebase and update the PR

steinaol (approved):
LGTM

kpmeen (approved):

PR #281, opened 2017-03-27
Musark 927 Add organisation with synonyms and serviceTags - service-actor

ellenjo:
Add method GetByNameAndTags that will return a list of organisations with the right serviceTag. It's the organisationService.find-method that use it.
Changed the two tables organization and organization_address to organisation and organisation_address.


musit-service/src/test/scala/no/uio/musit/service/MusitSearchSpec.scala
```diff
@@ -47,6 +47,13 @@ class MusitSearchSpec extends WordSpec with MustMatchers {
       MusitSearch.parseSearch("") mustBe MusitSearch.empty
     }
 
+    "tada" in {
```
kpmeen: This test needs to be properly described

service_actor/test/repositories/dao/AddressDaoSpec.scala
```diff
@@ -85,18 +86,19 @@ class AddressDaoSpec
         res1.addressType mustBe "WORK2"
         res1.streetAddress mustBe "Adressen2"
         res1.postalCode mustBe "0122"
-        res1.id mustBe Some(DatabaseId(3))
+        res1.id mustBe Some(DatabaseId(22))
 
-        val orgUpd = Organisation(
+        /* val orgUpd = Organisation(
```
kpmeen: Remove unused code

kpmeen (changes requested):
Just a couple of minor cleanups that you should do :-)

kpmeen (approved):

PR #285, opened 2017-03-30
MUSARK-956

ellenjo:
I have two comments on the add test because I can't find out about dateTime. I will test if the event has the same registeredDate and EventDate as the SampleObject.registeredDate.

### DIFF
--- a/service_management/app/models/analysis/events/Analysis.scala
+++ b/service_management/app/models/analysis/events/Analysis.scala
@@ -106,7 +106,6 @@ object AnalysisEvent extends WithDateTimeFormatters {
         Json.obj(tpe -> AnalysisCollection.discriminator)
 
     case sc: SampleCreated =>
-      implicit val scw = SampleCreated.writes
       SampleCreated.writes.writes(sc).as[JsObject] ++
         Json.obj(tpe -> SampleCreated.discriminator)
   }
@@ -209,10 +208,8 @@ object SampleCreated extends WithDateTimeFormatters {
 
   val discriminator = "SampleCreated"
 
-  def reads(implicit r: Reads[Analysis]): Reads[SampleCreated] =
-    Json.reads[SampleCreated]
+  val reads: Reads[SampleCreated] = Json.reads[SampleCreated]
 
-  def writes(implicit w: Writes[Analysis]): Writes[SampleCreated] =
-    Json.writes[SampleCreated]
+  val writes: Writes[SampleCreated] = Json.writes[SampleCreated]
 
 }
--- a/service_management/app/models/analysis/events/Analysis.scala
+++ b/service_management/app/models/analysis/events/Analysis.scala
@@ -33,6 +33,7 @@ sealed trait AnalysisEvent {
     this match {
       case a: Analysis            => a.copy(result = res)
       case ac: AnalysisCollection => ac
+      case sc: SampleCreated      => sc
     }
   }
 
@@ -44,6 +45,7 @@ sealed trait AnalysisEvent {
     this match {
       case a: Analysis            => Some(a.copy(result = res))
       case ac: AnalysisCollection => None
+      case sc: SampleCreated      => None
     }
 }
 
@@ -62,6 +64,7 @@ object AnalysisEvent extends WithDateTimeFormatters {
   implicit val reads: Reads[AnalysisEvent] = Reads { jsv =>
     implicit val ar  = Analysis.reads
     implicit val acr = AnalysisCollection.reads
+    implicit val sc  = SampleCreated.reads
 
     (jsv \ tpe).validateOpt[String] match {
       case JsSuccess(maybeType, path) =>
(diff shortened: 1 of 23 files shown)


service_management/app/models/analysis/events/Analysis.scala
```diff
@@ -173,3 +184,35 @@ object AnalysisCollection extends WithDateTimeFormatters {
...
+  val completedDate: Option[DateTime] = None
+  val analysisTypeId                  = SampleCreated.sampleEventTypeId
+}
+
+object SampleCreated extends WithDateTimeFormatters {
+  val sampleEventTypeId: AnalysisTypeId = AnalysisTypeId.empty
+
+  val discriminator = "SampleCreated"
+
+  def reads(implicit r: Reads[Analysis]): Reads[SampleCreated] =
```
kpmeen: As it stands, the _input_ argument has the wrong type.

When above comment is applied, this needs to change into an `implicit val reads: Reads[SampleCreated]` that also handles the `discriminator` as handled in the current code in the `AnalysisEvent` companion object.

service_management/app/models/analysis/events/Analysis.scala
```diff
@@ -173,3 +184,35 @@ object AnalysisCollection extends WithDateTimeFormatters {
...
+    Json.reads[SampleCreated]
+
+  def writes(implicit w: Writes[Analysis]): Writes[SampleCreated] =
```
kpmeen: As it stands, the _input_ argument has the wrong type.

When above comment is applied, this needs to change into an `implicit val writes: Writes[SampleCreated]` that also applies the `discriminator` as handled in the current code in the `AnalysisEvent` companion object.

service_management/app/repositories/analysis/dao/Tables.scala
```diff
@@ -145,33 +147,36 @@ trait Tables extends HasDatabaseConfigProvider[JdbcProfile] with ColumnTypeMappe
...
-    val container          = column[Option[String]]("SAMPLE_CONTAINER")
-    val storageMedium      = column[Option[String]]("STORAGE_MEDIUM")
-    val note               = column[Option[String]]("NOTE")
-    val registeredBy       = column[Option[ActorId]]("REGISTERED_BY")
-    val registeredDate     = column[Option[JSqlTimestamp]]("REGISTERED_DATE")
-    val updatedBy          = column[Option[ActorId]]("UPDATED_BY")
-    val updatedDate        = column[Option[JSqlTimestamp]]("UPDATED_DATE")
+    val id       = column[ObjectUUID]("SAMPLE_UUID", O.PrimaryKey)
+    val parentId = column[Option[ObjectUUID]]("PARENT_OBJECT_UUID")
+    //val isCollectionObject = column[Boolean]("IS_COLLECTION_OBJECT")
```
kpmeen: Remove commented out code

service_management/app/services/analysis/AnalysisService.scala
```diff
@@ -66,6 +70,17 @@ class AnalysisService @Inject()(
     analysisDao.insertCol(acol)
   }
 
+  /* def addSampleCreated(
```
kpmeen: Remove commented out code.

service_management/test/resources/evolutions/test/1.sql
```diff
@@ -5,7 +5,9 @@ CREATE SCHEMA IF NOT EXISTS MUSARK_ANALYSIS;
 CREATE TABLE MUSARK_ANALYSIS.SAMPLE_OBJECT (
   sample_uuid          VARCHAR2(36)      NOT NULL,
   parent_object_uuid   VARCHAR2(36),
-  is_collection_object INTEGER DEFAULT 0 NOT NULL,
+  --is_collection_object INTEGER DEFAULT 0        NOT NULL,
```
kpmeen: remove commented out code

service_management/test/resources/logback-test.xml
```diff
@@ -13,7 +13,7 @@
 
   <logger name="play" level="WARN"/>
   <logger name="akka" level="ERROR"/>
-  <logger name="slick.jdbc" level="WARN"/>
+  <logger name="slick.jdbc" level="DEBUG"/>
```
kpmeen: Please change back to `WARN` before committing and pushing code.
It's a good idea to get used to looking over each file you want to commit and adding them explicitly, instead of doing a `git add .` from the root of the project.

kpmeen (changes requested):

service_management/test/services/analysis/SampleObjectServiceSpec.scala
```diff
@@ -1,3 +1,85 @@
...
+          parentId = ObjectUUID.generateAsOpt(),
+          isExtracted = true
+        )
+      val res  = service.add(so).futureValue
+      val find = service.findById(res.get.underlying).futureValue
+      find.successValue.value.isExtracted mustBe true
+      val oid   = find.successValue.value.parentObjectId
+      val event = eventService.findByObject(oid.value).futureValue.successValue
+      event.size mustBe 1
+      event.head.analysisTypeId mustBe SampleCreated.sampleEventTypeId
```
ingarabr: Take a closer look at scalatest and you will find some matchers that remove the call to the head of the collection.

kpmeen (approved):

PR #288, opened 2017-03-31
MUSARK-572: Refactoring service_storagefacility

kpmeen:
The end is nigh for ye olde EventDao and Event DTOs. This PR does _not_ change the API's in any way. That will have to come in a separate PR later. The intention here, is to ensure that we are able to migrate from old data types/tables to new ones. We need to know how the migration code behaves on the UTV environment with ~900K events ASAP.

There is a chance that the current migration implementation is a bit heavy on the memory usage, meaning we'll have to implement this in a slightly different way.

But...this is the easiest and quickest alternative implement, and if it works...yay. It's a one off job and the code can be removed once successfully applied.


kpmeen:
Seems, with 90% certainty, that the current implementation won't behave well against the size of real-world data. So I'll need to change loading of old data from regular `DBIO` to`StreamingDBIO`. The migration part will then use the Akka Streams API to feed data back into the DB.

kpmeen:
The code should be ready for review now. Don't bother to much about the stuff related to the migration of old to new events.

There's a huge amount of code that can be removed once the API starts using the new services and repositories for events. But for now, we have to live with a bit of clutter.

musit-models/src/main/scala/no/uio/musit/functional/MonadTransformers.scala
```diff
@@ -53,6 +55,25 @@ object MonadTransformers {
...
+    /**
+     * Returns a MusitResultT[Future, A] where the Future is successfully
+     * evaluated with the given MusitResult[A] argument.
+     *
+     * @param res the MusitResult to wrap
+     * @param m an implicitly provided Monad[Future]
+     * @tparam A the type we want to access inside the transformer functions
+     * @return {{{MusitResultT[Future, A]}}}
+     */
+    def successful[A](
```
ingarabr, line 69: Should we take some more inspiration from for instance this post: http://blog.leifbattermann.de/2017/03/16/7-most-convenient-ways-to-create-a-future-either-stack/

It might make it easier to work with `Future` and `MusitResult`s. It's out of the scope for this PR but I think it is something we should look into.

service_storagefacility/app/migration/EventMigrator.scala
```diff
@@ -0,0 +1,249 @@
...
+        convertAndInsert(curr).map { res =>
+          acc.copy(
+            success = acc.success + res.success,
+            errors = acc.errors ++ res.errors
+          )
+        }
+      }
+      .map { res =>
+        val endTime   = System.currentTimeMillis() milliseconds
+        val totalTime = endTime - startTime
```
ingarabr, line 197: I have implemented a stop watch that's located in `musit-service`

service_storagefacility/app/models/storage/event/control/Control.scala
```diff
@@ -0,0 +1,40 @@
...
+import models.storage.event.control.ControlAttributes._
+import models.storage.event.{EventType, MusitEvent}
+import no.uio.musit.formatters.WithDateTimeFormatters
+import no.uio.musit.models.{ActorId, EventId, StorageNodeId}
+import org.joda.time.DateTime
+import play.api.libs.json._
+
+/**
+ * TODO: Document me
+ */
```
ingarabr, line 12: Do you use a template with the todo document me?

service_storagefacility/app/models/storage/event/dto/DtoConverters.scala
```diff
@@ -684,8 +661,19 @@ object DtoConverters {
...
     )(init: (EventType, Option[StorageNodeDatabaseId], StorageNodeDatabaseId) => A): A = {
       val eventType = EventType.fromEventTypeId(dto.eventTypeId)
-      val from      = dto.valueLong
-      val to        = dto.relatedPlaces.head.placeId
+      val from: Option[StorageNodeDatabaseId] =
+        dto.valueLong
+          .map(StorageNodeDatabaseId.apply)
+          .orElse(dto.relatedPlaces.find(_.roleId == 4).map(_.placeId))
+      val to = dto.relatedPlaces
+        .find(rid => rid.roleId == 3 || rid.roleId == 1)
```
ingarabr, line 669: Magic numbers related to roleId.

service_storagefacility/app/models/storage/event/envreq/EnvRequirement.scala
```diff
@@ -0,0 +1,92 @@
...
+    hypoxicAir: Option[Interval[Double]],
+    cleaning: Option[String],
+    light: Option[String]
+) extends MusitEvent {
+
+  override type T = EnvRequirement
+
+  override def withId(id: Option[EventId]) = copy(id = id)
+
+  def similar(er: EnvRequirement): Boolean = {
```
ingarabr, line 32: Not related to this PR but this feels like a code smell.  If we have grouped all the environment data into a case class we could have just ensure that the case classes is similar.

service_storagefacility/app/models/storage/event/observation/ObservationAttributes.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  case class ObservationCleaning(
+      note: Option[String],
+      cleaning: Option[String]
+  ) extends ObservationAttribute
+
+  object ObservationCleaning {
+    implicit val formats: Format[ObservationCleaning] = (
+      (__ \ "note").formatNullable[String](maxCharsFormat(250)) and
+        (__ \ "cleaning").formatNullable[String](maxCharsFormat(250))
```
ingarabr, line 76: We need to find a way to format this better. Had been nice if it has been align under each other

service_storagefacility/app/repositories/storage/dao/MigrationDao.scala
```diff
@@ -0,0 +1,414 @@
...
+      Option[String],
+      Option[Long],
+      Option[Long],
+      Option[String],
+      Option[Double],
+      Option[String],
+      Option[JSqlTimestamp],
+      // actors
+      Option[Int],
+      Option[String],
```
ingarabr, line 51: Something I considered when I hit the limit of the tuple size. The way your grouping actor, places etc can be done by wrapping them into a tuple. Then it also will be a bit easier to use in the mappers.

service_storagefacility/app/repositories/storage/dao/nodes/BuildingDao.scala
```diff
@@ -103,13 +98,7 @@ class BuildingDao @Inject()(
...
 
       }
-      .recover {
-        case NonFatal(ex) =>
-          val msg = s"There was an error updating building $id"
-          logger.debug(s"Using $id, building has ID ${building.id}")
-          logger.error(msg, ex)
-          MusitDbError(msg, Some(ex))
-      }
+      .recover(nonFatal(s"There was an error updating building $id"))
```
ingarabr, line 101: I like this change :)

service_storagefacility/app/repositories/storage/old_dao/event/EventDao.scala
```diff
@@ -579,6 +578,33 @@ class EventDao @Inject()(
...
+          .map {
+            case MusitSuccess(ms) =>
+              ms.map(dto => success(dto))
+
+            case err: MusitError =>
+              logger.error(err.message)
+              throw new IllegalStateException("Impossible state") // scalastyle:ignore
+          }
+          .filter(_.isDefined)
+          .map(_.get)
```
ingarabr, line 602: This feels wrong to do.

musit-models/src/main/scala/no/uio/musit/functional/MonadTransformers.scala
kpmeen: I've considered it. But, when considering all the factors at play, I've come to the conclusion that it's  not necessarily a good idea.

We previously had something similar to what's described in that blog post. But, it caused more confusion than necessary. And it was poorly implemented. Being explicit about what's happening helped the understanding of what's going on.

service_storagefacility/app/models/storage/event/control/Control.scala
kpmeen: No...this was added way back when...and the code in the "new" packages are copied from the "old" package. Only updating some of the member fields.

service_storagefacility/app/models/storage/event/dto/DtoConverters.scala
kpmeen: Yes...this is old code that will disappear.

service_storagefacility/app/migration/EventMigrator.scala
kpmeen: cool...didn't know about that. Not too worried about that for this piece of code though. It'll be deleted in a release or two :-)

service_storagefacility/app/repositories/storage/dao/MigrationDao.scala
kpmeen: Yes... in this case though, I'm only reading out what I need for the migration. Initially I was reading out more than I actually needed and had a 25-uple :-p

service_storagefacility/app/repositories/storage/old_dao/event/EventDao.scala
kpmeen: Yes...in any other scenario this would be a no-go. But here it's actually just convenience for the migration. After migration this entire file will be removed. So I wouldn't be too concerned about it.

service_storagefacility/app/models/storage/event/observation/ObservationAttributes.scala
kpmeen: It's continuation indentation...quite standard. Not going to change that.

musit-models/src/main/scala/no/uio/musit/functional/MonadTransformers.scala
ingarabr: Yeah, if we're taking that path we have to find the balance so it's not to much black magic for the less experiences developers.

service_storagefacility/app/migration/EventMigrator.scala
ingarabr: You have to take a closer look at the PRs I give you ;-)

service_storagefacility/app/models/storage/event/dto/DtoConverters.scala
ingarabr: 👍

ingarabr (approved):

PR #290, opened 2017-04-05
MUSARK-979: service endpoints for updating analysis events and results

kpmeen:
Implements service endpoints for updating analysis events and results.

Important note: This PR requires an update to the oracle table `MUSARK_ANALYSIS.RESULT` and `MUSARK_ANALYSIS.EVENT`.

Another note:
@rituvesh when this PR is merged, you should change the client to _always_ send `AnalysisCollections` when creating a new analysis. Regardless of it being one or more objects associated with it.

API for updating an analysis is the same as when creating it. Updating a result also has the same message format as adding a result.


service_management/test/services/analysis/SampleObjectServiceSpec.scala
```diff
@@ -56,40 +58,50 @@ class SampleObjectServiceSpec
...
       registeredDate = Some(now),
       updatedBy = None,
       updatedDate = None
     )
   }
 
   "The SampleObjectService " should {
-    "returns the sampleObject that is inserted" in {
+    val parentId                    = ObjectUUID.generate()
+    var addedId: Option[ObjectUUID] = None
```
kpmeen, line 70: The mutable field is required here to be able to get a hold of the ObjectUUID

service_management/app/repositories/analysis/dao/AnalysisDao.scala
```diff
@@ -73,7 +86,9 @@ class AnalysisDao @Inject()(
 
     query.result.map { res =>
       res.map { row =>
-        fromEventRow(row._1).flatMap(_.withResultAsOpt(fromResultRow(row._2))).get
+        toAnalysis(row._1)
```
kpmeen, line 89: necessary

service_management/app/repositories/analysis/dao/AnalysisDao.scala
```diff
@@ -131,6 +146,42 @@ class AnalysisDao @Inject()(
...
+   *
+   * @param id the EventId associated with the analysis event to update
+   * @param ae the AnalysisEvent to update
+   * @return a result with an option of the updated event
+   */
+  def update(
+      id: EventId,
+      ae: AnalysisEvent
+  ): Future[MusitResult[Option[AnalysisEvent]]] = {
+    val action = updateAction(id, ae).transactionally
```
ingarabr, line 160: I prefer that we change name of the `updateAction` function to indicate that it update by the `id`. Also I think that will make the flatmap code below a bit easer to understand sine we can/should only have zero or one updated row. The reason for it is that I thought about the case where it was higher than 1 and if we should rollback the update, something that should  never happen.

service_management/app/repositories/analysis/dao/AnalysisDao.scala
```diff
@@ -190,21 +241,51 @@ class AnalysisDao @Inject()(
...
+   * @param res The AnalysisResult to add.
+   * @return eventually a result with the EventId the saved result belongs to.
+   */
+  def upsertResult(id: EventId, res: AnalysisResult): Future[MusitResult[EventId]] = {
+    val action = upsertResultAction(id, res)
+
+    db.run(action.transactionally)
+      .flatMap { numUpdated =>
+        val q = resultTable.filter(_.eventId === id).map(_.eventId)
+        db.run(q.result.headOption).map {
```
ingarabr, line 273: Had it been cleaner if we used a for comprehension for both actions here? Or do you explicitly want the update even if the query for the result to fails?

service_management/test/repositories/analysis/dao/AnalysisDaoSpec.scala
```diff
@@ -61,52 +28,56 @@ class AnalysisDaoSpec
...
     "fetching analysis events" should {
 
-      "return an analysis collection with children" in {
+      "return an analysis collection with many children" in {
         val res = dao.findById(EventId(2)).futureValue
 
         res.successValue.value mustBe an[AnalysisCollection]
         val ac = res.successValue.value.asInstanceOf[AnalysisCollection]
         ac.events must not be empty
+        ac.events.size mustBe 3
```
ingarabr, line 69: `ac.events must have size(3)` and remove the empty assert
goes for all the asserts with that combo.

service_management/test/repositories/analysis/dao/AnalysisDaoSpec.scala
```diff
@@ -178,9 +139,90 @@ class AnalysisDaoSpec
...
+      "successfully add a result to an analysis" in {
+        val e1 = dummyAnalysis(Some(oid1))
+        val ac = dummyAnalysisCollection(None, e1)
+        val gr = dummyGenericResult(comment = Some("updated result"))
+
+        val eid = dao.insertCol(ac).futureValue.successValue
+
+        dao.upsertResult(eid, gr).futureValue.isSuccess mustBe true
+
+        dao.findById(eid).futureValue.successValue.value match {
```
ingarabr, line 154: Another way to do this is to use the Inside trait.

service_management/app/repositories/analysis/dao/AnalysisDao.scala
kpmeen: The update in these scenarios are _always_ by id. So the naming it updateByIdAction seems redundant to me.

service_management/app/repositories/analysis/dao/AnalysisDao.scala
kpmeen: Because it's an "upsert", the only time the "update" will fail is if there's a DB exception (which will cause a Future.failed). If the result doesn't exist, it will be added, if it exists it will be updated.
With only 2 levels deep blocks, for-comprehensions vs flatMap map... doesn't really matter all that much. Depends on the granularity of control you want.

service_management/test/repositories/analysis/dao/AnalysisDaoSpec.scala
kpmeen: That's a legitimate point. However, since the above code is consistent with the rest of our codebase, it's not a grounds for not merging it.

service_management/test/repositories/analysis/dao/AnalysisDaoSpec.scala
kpmeen: yes....that's a matter of taste really.

PR #291, opened 2017-04-06
fix: return 503 when service is not available in service_geo_location

ellenjo:
I didn't make a test in controller for this issue...

### DIFF
--- a/service_geo_location/app/controllers/GeoLocationController.scala
+++ b/service_geo_location/app/controllers/GeoLocationController.scala
@@ -48,14 +48,11 @@ class GeoLocationController @Inject()(
     geoLocService
       .searchGeoNorway(expression)
       .map {
-        case MusitSuccess(locations) => {
+        case MusitSuccess(locations) =>
           Ok(Json.toJson(locations))
-        }
 
-        case MusitHttpError(status, msg) => {
-          println("inni map. MusitHttpError")
+        case MusitHttpError(status, msg) =>
           Status(status)
-        }
 
         case err: MusitError =>
           logger.error("InternalServerError: " + s"${err.message}")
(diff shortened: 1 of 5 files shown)


service_geo_location/app/controllers/GeoLocationController.scala
```diff
@@ -46,8 +47,19 @@ class GeoLocationController @Inject()(
     val expression = search.getOrElse("")
     geoLocService
       .searchGeoNorway(expression)
-      .map { location =>
-        Ok(Json.toJson(location))
+      .map {
+        case MusitSuccess(locations) => {
```
kpmeen: Remove unnecessary `{` and `}` after block starting with `=>`. This applies to all `case` statements below.

service_geo_location/test/controllers/GeoLocationControllerIntegrationSpec.scala
```diff
@@ -75,7 +84,10 @@ class GeoLocationControllerIntegrationSpec extends MusitSpecWithServerPerSuite {
         (jsArr.head \ "place").as[String] mustBe "OSLO"
         (jsArr.head \ "zip").as[String] mustBe "0192"
       }
+
+      "return musitHttpError if service not available" in {
+        // todo
```
kpmeen: Instead of `// todo` use `pending` from ScalaTest

```scala
"return musitHttpError if service not available" in {
  pending
}
```

kpmeen: But, you can actually just remove this empty test. The test scenario in the `GeoLocationServiceSpec` should be enough.

kpmeen (changes requested):

kpmeen (approved):

PR #294, opened 2017-04-11
Run tests from intellij

ingarabr:
Intellij Idea doesn't read the sbt config when running tests
from the editor. All tests that's not depending on configuration
loaded from the application.test.conf does work without any
modification. This is a workaround that enables us to run all the tests
trough Idea in a way that doesn't conflict with how sbt setup the test
environment.

https://www.playframework.com/documentation/2.5.x/ProductionConfiguration

### DIFF
--- a/musit-test/src/main/scala/no/uio/musit/test/MusitFakeApplication.scala
+++ b/musit-test/src/main/scala/no/uio/musit/test/MusitFakeApplication.scala
@@ -4,7 +4,17 @@ import play.api.inject.guice.GuiceApplicationBuilder
 
 trait MusitFakeApplication extends TestConfigs {
 
-  def createApplication() =
+  def createApplication() = {
+    if (notExecutedFromSbt) {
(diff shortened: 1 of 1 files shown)


musit-test/src/main/scala/no/uio/musit/test/MusitFakeApplication.scala
```diff
@@ -1,29 +1,20 @@
...
-  def createApplication() =
+  def createApplication() = {
+    if (notExecutedFromSbt) {
+      setProperty("config.file", "/application.test.conf")
+      setProperty("logger.file", "/logback-test.xml")
+    }
     new GuiceApplicationBuilder().configure(slickWithInMemoryH2()).build()
+  }
+
+  private def notExecutedFromSbt = System.getProperty("config.file") == null
```
jarlah: Scala.sys?

jarlah (approved):
looks good as far as  I can see

PR #295, opened 2017-04-11
Musark 933 - Adding module to service_auth

jarlah:
This issue/pr is a work in progress. I made a PR  of it, to make any big changes discussable.

TODOS:
* test DONE
* update group management gui to include support for selecting module DONE
* consider refactor tables, if relevant NOT RELEVANT (?)


musit-models/src/main/scala/no/uio/musit/models/Module.scala
```diff
@@ -0,0 +1,25 @@
...
+  val id: Int
+}
+
+case object StorageFacility extends Module {
+  val id = 1
+}
+case object LoanProcess extends Module {
+  val id = 2
+}
+
```
kpmeen: `LoanProcess` can be renamed to `Loan` since that's the business' name for the module.
Also we need an `Analysis` module.

musit-models/src/main/scala/no/uio/musit/models/Module.scala
```diff
@@ -0,0 +1,25 @@
...
+object Module {
+
+  implicit val reads: Reads[Module]   = __.read[Int].map(fromInt)
+  implicit val writes: Writes[Module] = Writes{m => JsNumber(m.id)}
+
+  def fromInt(id: Int): Module = id match {
+    case StorageFacility.id => StorageFacility
```
kpmeen: This function is incomplete in the match since it doesn't account for arguments that doesn't match any of the predefined Modules. Return type should then also be `Option[Module]`

If you need a function that should return a non Option Module, you still need to account for ID's that doesn't match.
The function is then `unsafe` and should be defined as follows:

```scala
@throws(classOf[NoSuchElementException])
def unsafeFromInt(id: Int): Module = fromInt(id).get
```
For a similar example, see `MusitUUIDOps`.

musit-models/src/main/scala/no/uio/musit/models/Module.scala
```diff
@@ -0,0 +1,25 @@
+package no.uio.musit.models
+
+import play.api.libs.json.{JsNumber, Reads, Writes, __}
+
+sealed trait Module {
```
ingarabr: Can we rename this to GroupModule or MusitModule (or so) to not get name conflicts with for instance guice modules.

musit-service/src/main/scala/no/uio/musit/security/AuthTables.scala
```diff
@@ -80,6 +80,12 @@ trait AuthTables extends HasDatabaseConfigProvider[JdbcProfile] with DateTimeImp
       i => MuseumId.fromInt(i)
     )
 
+  implicit lazy val moduleMapper: BaseColumnType[Module] =
+    MappedColumnType.base[Module, Int](
+      m => m.id,
+      i => Module.fromInt(i)
```
kpmeen: When above comment is implemented. This should be `Module.unsafeFromInt(i)`

kpmeen (changes requested):

kpmeen:
A more comprehensive review when there's more to review :-)

musit-models/src/main/scala/no/uio/musit/models/Module.scala
```diff
@@ -0,0 +1,25 @@
...
+  val id: Int
+}
+
+case object StorageFacility extends Module {
+  val id = 1
+}
+case object LoanProcess extends Module {
```
ingarabr: Remove, we don't have any requirements atm for loans.

service_auth/app/models/Group.scala
```diff
@@ -53,15 +57,17 @@ object GroupAdd {
 
   implicit def reads: Reads[GroupAdd] = Json.reads[GroupAdd]
 
-  def applyForm(name: String, permInt: Int, mid: Int, maybeDesc: Option[String]) =
-    GroupAdd(name, Permission.fromInt(permInt), MuseumId(mid), maybeDesc)
+  def applyForm(name: String, m: Int, permInt: Int, mid: Int, maybeDesc: Option[String]) =
```
ingarabr: give the `m` a better name since it's only a int type.

musit-models/src/main/scala/no/uio/musit/models/Module.scala
jarlah: The "LoanProcess" type was meant as irony. But it needs a little bit of context ;) But yes, it should be Loan.

musit-models/src/main/scala/no/uio/musit/models/Module.scala
jarlah: I am not very convinced about this, at all. All the rows in the db column GROUP_MODULE will either be set by migration script or validated in through the form controllers for the group admin gui. So I really dont see why we should need to wrap an exception that will never happen. And exception if it ever occurs will be non exhaustive match exception given the excact line number and place in the code. This will tell us whats the problem. We dont need NoSuchElementException for that. Its overengineering

musit-models/src/main/scala/no/uio/musit/models/Module.scala
jarlah: AccessModule or sumtin. But yes. Well rename it

musit-models/src/main/scala/no/uio/musit/models/Module.scala
jarlah: It will come soon. It does no harm. But I can rename it.

service_auth/app/models/Group.scala
jarlah: Agree

musit-models/src/main/scala/no/uio/musit/models/Module.scala
kpmeen: Database mapper is only one point in the code where `fromInt` is used. There's also the implicit reads for Module.
We cannot 100% control the quality of data being sent into the API. And handling the parsing correctly is essential for correctness and being able to provide a sensible error message to the client.

The `@throws` annotation is for clarity that the `unsafeFromInt` function is just that, unsafe. It is the `get` operation on the `Option` that throw the actual Exception. If you want to be explicit about what kind of exception you want to throw, an `IllegalArgumentException` would be OK.


correct implicit reads implementation
```scala
implicit val reads: Reads[Module] = Reads { jsv =>
  jsv.validate[Int] match {
    case JsSuccess(id, path) =>
      fromInt(id).map(JsSuccess.apply).getOrElse(JsError(path, s"Unknown Module ID $id"))

    case err: JsError =>
      err
  }
}
```

So please change as initially described. (Including the `reads` from this comment)

musit-models/src/main/scala/no/uio/musit/models/GroupModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+      case err: JsError =>
+        err
+    }
+  }
+
+  implicit val writes: Writes[GroupModule] = Writes{ m => JsNumber(m.id)}
+
+  @throws(classOf[NoSuchElementException])
+  def unsafeFromInt(id: Int): GroupModule = fromInt(id).get
```
jarlah: When I program in Java, I tend to avoid catching null pointer exceptions that should never occur, because, they should never occur. Our discussion above, centers around the fact that there is not a single endpoint that allows you to post an invalid moduleId. If its invalid, its rejected. Therefore I still think its unnecessary with an unsafeFromInt, in general, due to the fact that what we are actually guarding against is not actually invalid ids, because thats impossible, but we are guarding against debugging a hard to understand stacktrace that tells us that a value fell through a case match.

jarlah: But its not important for me :) I am just stating my loose opinions on the subject. Which, given the complexity of the code we are talking about, is pretty useless really ;)

service_auth/app/models/Group.scala
```diff
@@ -53,15 +57,17 @@ object GroupAdd {
...
-  def unapplyForm(g: GroupAdd) =
-    Some((g.name, g.permission.priority, g.museumId.underlying, g.description))
+  def unapplyForm(g: GroupAdd) = Some(
+    (g.name, g.module.id, g.permission.priority, g.museumId.underlying, g.description)
+  )
 
   val groupAddForm = Form(
     mapping(
       "name"        -> text(minLength = 3),
+      "module"      -> number.verifying(GroupModule.fromInt(_).nonEmpty),
```
ingarabr, line 67: This will break the admin gui.

jarlah: This is now fixed, by adding an extra dropdown in the backend interface for group management

musit-service/src/main/scala/no/uio/musit/security/AuthenticatedUser.scala
```diff
@@ -62,8 +63,9 @@ case class AuthenticatedUser(
       logger.debug(s"User with GodMode accessing system.")
       MusitSuccess(())
     } else if (isAuthorizedFor(museum)) {
-      val allowed = permissionsFor(museum).exists(_.priority >= lowest.priority)
-      if (allowed) MusitSuccess(())
+      val allowedForPermissions = permissionsFor(museum).exists(_.priority >= lowest.priority)
+      val allowedForModule = module.foldLeft(true)((_, m) => this.groups.exists(_.module == m))
```
ingarabr: Can you rewrite this to use pattern matcher. Easer to see what you're trying to achieve then. :)

musit-service/src/main/scala/no/uio/musit/security/AuthTables.scala
```diff
@@ -80,6 +80,12 @@ trait AuthTables extends HasDatabaseConfigProvider[JdbcProfile] with DateTimeImp
       i => MuseumId.fromInt(i)
     )
 
+  implicit lazy val moduleMapper: BaseColumnType[GroupModule] =
```
ingarabr: groupModuleMapper

musit-models/src/main/scala/no/uio/musit/models/GroupModule.scala
```diff
@@ -0,0 +1,37 @@
+package no.uio.musit.models
+
+import play.api.libs.json._
+
+sealed trait GroupModule {
```
kpmeen, line 5: I think one of the reasons why I don't like this name is because it's a bit too easy to confuse this with a  guide module. Which, by convention, are named `*Module`. But that's the domain language, so hard to do anything about.

The main problem here is that this type doesn't specifically have anything to do with a group. It only describes a module in the MUSIT system. So the trait should be renamed to just `Module` or even `MusitModule`. File should be renamed accordingly.

musit-service/src/main/scala/no/uio/musit/security/AuthenticatedUser.scala
```diff
@@ -62,8 +63,13 @@ case class AuthenticatedUser(
       logger.debug(s"User with GodMode accessing system.")
       MusitSuccess(())
     } else if (isAuthorizedFor(museum)) {
-      val allowed = permissionsFor(museum).exists(_.priority >= lowest.priority)
-      if (allowed) MusitSuccess(())
+      val allowedForPermissions = permissionsFor(museum).exists { perm =>
+        perm.priority >= lowest.priority
+      }
+      val allowedForModule = module.foldLeft(true) {
```
kpmeen, line 69: `val allowedForModule = module.exists(m => groups.exists(_.module == m))`

service_auth/test/repositories/dao/AuthDaoSpec.scala
```diff
@@ -87,7 +87,19 @@ class AuthDaoSpec
...
+            msg must include("An error occurred")
+            ex must not be None
+
+          case err =>
+            fail("Expected MusitDbError")
+        }
+      }
+
+      "fail if the module is null" in {
+        val grp = GroupAdd("testFail", null, Permissions.Read, Test.id, Some("test group fail")) // scalastyle:ignore
```
kpmeen, line 102: We need a value to insert for users that have e.g. `GodMode` privileges. For the `Museums` we have an `All` that will be inserted in case of the all mighty. Should possibly add an `All` for `Module` as well.

musit-models/src/main/scala/no/uio/musit/models/GroupModule.scala
```diff
@@ -0,0 +1,37 @@
...
+  val id: Int
+}
+
+case object StorageFacility extends GroupModule { val id = 1 }
+case object Loan extends GroupModule { val id = 2 }
+case object Analysis extends GroupModule { val id = 3 }
+
```
kpmeen, line 12: for `GodMode` (and other super users, e.g. `MusitAdmin`) it's good to have an `All` here as well.

service_auth/test/resources/evolutions/test/1.sql
```diff
@@ -55,6 +56,6 @@ INSERT INTO MUSARK_AUTH.MUSEUM_COLLECTION VALUES ('ba3d4d30-810b-4c07-81b3-37751
 INSERT INTO MUSARK_AUTH.MUSEUM_COLLECTION VALUES ('ef4dc066-b6f8-4155-89f8-7aa9aeeb2dc4', 'Marine evertebrater', '[10]');
 INSERT INTO MUSARK_AUTH.MUSEUM_COLLECTION VALUES ('c39703f4-c64c-4c13-b92c-67e6af68f674', 'ArkEtno', '[1,2]');
 
-INSERT INTO MUSARK_AUTH.AUTH_GROUP VALUES ('7963557f-e36b-4f12-951d-daa36b3840e1', 'test_test', 20, 99, 'Foo bar baz');
+INSERT INTO MUSARK_AUTH.AUTH_GROUP VALUES ('7963557f-e36b-4f12-951d-daa36b3840e1', 'test_test', 1, 20, 99, 'Foo bar baz');
```
kpmeen, line 59: Just a thought, so we don't forget. We should probably provide a SQL script that updates all the _currently_ registered users to have access to storage facility. Unless the user has super user access, where we should insert the value for the `All` "module".

kpmeen (changes requested):

PR #297, opened 2017-04-21
MUSARK-993 - feature/missing sample fields

ingarabr:
- treatment
- residualMaterial
- description
- externalId.source


service_management/app/models/analysis/ResidualMaterials.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object ResidualMaterials {
+
+  sealed trait ResidualMaterial {
+    val identity: Int
+  }
+
+  object ResidualMaterial {
+    def fromInt(i: Int): Option[ResidualMaterial] = i match {
+      case NotSpecified.identity        => Some(NotSpecified)
```
kpmeen: It's preferable give `identity` a different name. So there's no ambiguity with `Predef.identity`.

service_management/app/repositories/analysis/dao/Tables.scala
```diff
@@ -304,7 +309,7 @@ trait Tables extends HasDatabaseConfigProvider[JdbcProfile] with ColumnTypeMappe
       maybeTuple: Option[ResultRow]
   ): Option[AnalysisResult] =
     maybeTuple.flatMap(fromResultRow)
-
+  //todo add externalIdSource, treatment, residualMaterial, description
```
kpmeen: Is this comment still valid?

kpmeen (commented):
Just one mine thing

service_management/app/models/analysis/ResidualMaterials.scala
ingarabr: nope, I'll remove it

service_management/app/repositories/analysis/dao/Tables.scala
ingarabr: nope, I'll remove it

steinaol: I have a question concerning different actors (persons) with different roles and dates connected to a sample (i.e. the sampleCreation [Prøveuttak]). From frontend we have an array with person-, role- and date objects and. How should we send this to backend? An array called Persons, where each entry is an object with type { name: string, role: string, date: string}?

ingarabr: Flatten them out based on there roles.

kpmeen (approved):

PR #306, opened 2017-04-28
Musark 904 - delete sample object

ellenjo:


service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
     soDao.listForMuseum(mid)
   }
 
+  def delete(
```
kpmeen, line 96: Let's look at what the function below does, exactly. But first, try to identify what we want to achieve.

The goal: We want to mark a sample object as deleted, and return a `MusitResult` that indicates a successful operation.

So, let's go through the function, line by line (where the interesting bits are).

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+      oid: ObjectUUID
+  )(implicit currUser: AuthenticatedUser): Future[MusitResult[Unit]] = {
+    def enrich(orig: SampleObject) = {
+      orig.copy(
+        isDeleted = true,
+        updatedStamp = Some(ActorStamp(currUser.id, dateTimeNow))
+      )
+    }
+
+    val updatedRes = for {
```
kpmeen, line 106: The for-comprehension is "where the magic happens", so let's focus on that.

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+      orig <- MusitResultT(findById(oid, MusitEmpty))
```
kpmeen, line 107: Step 1: Try to locate the `SampleObject` associated with the given `ObjectUUID`.

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+      _    <- MusitResultT(soDao.update(enrich(orig)))
```
kpmeen: Step 2: Modify the `isDeleted` and `updatedStamp` properties on the `SampleObject` that was found in the previous step (`orig`).

Step 3: Call the `SampleObjectDao.update` function to save the modified `SampleObject`.

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+      upd <- MusitResultT(
+              findById(
```
kpmeen: Step 4: Looking up the `SampleObject` again. In a "normal" update situation this is perfectly reasonable.

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+                oid,
+                MusitInternalError(s"Couldn't find sample $oid after delete")
+              )
+            )
+    } yield upd
```
kpmeen: The for-comprehension finally yields the updated `SampleObject` that was found.

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +93,29 @@ class SampleObjectService @Inject()(
...
+
+    // Need to do some tricks to align the shapes again.
+    updatedRes.value.map(_.map(_ => MusitSuccess(())))
```
kpmeen: The function finally maps into the `Future` and ignores the updated `SampleObject` value completely. Which is perfectly reasonable since it's been marked as deleted.

This tells us that the code might be doing more than it needs to. The unnecessary step in this case is Step 4.

Let's see how we can make the for-comprehension block a bit simpler:

```scala
def delete(... args ...) {
  ...
  val updatedRes = for {
    orig <- MusitResultT(findById(oid, MusitEmpty))
    del  <- MusitResultT(soDao.update(enrich(orig)).map(_ => MusitSuccess(()))
  } yield del

  updatedRes.value
}
```

Step 4 is now gone, as well as the last `map` function to transform `updatedRes`.

service_management/test/controllers/analysis/SampleObjectControllerIntegrationSpec.scala
```diff
@@ -188,6 +189,39 @@ class SampleObjectControllerIntegrationSpec
...
+      val expJs    = (all.json \ 2).as[JsObject]
+      val objectId = (expJs \ "objectId").as[String]
+
+      val res =
+        wsUrl(deleteUrl(mid)(objectId)).withHeaders(token.asHeader).get().futureValue
+
+      res.status mustBe OK
+    }
+
+    "try to delete a objectid that not exists, return NotFound" in {
```
kpmeen: this test description doesn't read very well when put into the surrounding context
```
"Invoking the sample object controller API" should "try to delete a objectid that not exists, return NotFound"
```

It sort of inverts the intention... a better phrasing could for example be

```
"Invoking the sample object controller API" should "return 404 when trying to delete a non-existing sample object"
```

service_management/test/controllers/analysis/SampleObjectControllerIntegrationSpec.scala
```diff
@@ -188,6 +189,39 @@ class SampleObjectControllerIntegrationSpec
...
+
+      val objectId = "123e4567-e89b-12d3-a456-426655440000"
+
+      val res =
+        wsUrl(deleteUrl(mid)(objectId)).withHeaders(token.asHeader).get().futureValue
+
+      res.status mustBe NOT_FOUND
+    }
+
+    "try to delete wrong objectid, return BadRequest" in {
```
kpmeen: A clearer test description here would be `"return 400 when trying to delete a sample object with an invalid UUID"`

kpmeen (changes requested):
The code is good @ellenjo. I did see a couple of things once I put on my review glasses. Mostly minor things :-)

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -105,17 +105,11 @@ class SampleObjectService @Inject()(
...
-                MusitInternalError(s"Couldn't find sample $oid after delete")
-              )
-            )
-    } yield upd
+      del  <- MusitResultT(soDao.update(enrich(orig))).map(_ => MusitSuccess(()))
+    } yield del
 
     // Need to do some tricks to align the shapes again.
-    updatedRes.value.map(_.map(_ => MusitSuccess(())))
+    updatedRes.value.map(_ => MusitSuccess(()))
```
kpmeen: on line 108 the expression giving a value to `del` is also explicitly mapped to `MusitSuccess(())`. So doing the same thing with `updatedRes.value` here shouldn't be necessary.

kpmeen (changes requested):

kpmeen (approved):

PR #307, opened 2017-04-28
MUSARK-1008: Refactoring API's to use the new backend in storagefacility

kpmeen:
This also has to include changes to `service_thingaggregate` where there are still some services that


service_storagefacility/app/controllers/storage/EventController.scala
```diff
@@ -48,20 +30,29 @@ class EventController @Inject()(
...
+          case JsError(errors) =>
+            Future.successful(BadRequest(JsError.toJson(errors)))
         }
-      case JsError(errors) =>
-        Future.successful(BadRequest(JsError.toJson(errors)))
-    }
+      }
+      .getOrElse {
+        Future.successful(
+          BadRequest(Json.obj("message" -> s"Invalid UUID $nodeId"))
```
ingarabr: This code block repeat itself a lot. Consider moving it to a function.

service_storagefacility/app/controllers/storage/StorageController.scala
```diff
@@ -334,14 +336,43 @@ final class StorageController @Inject()(
...
+  private def mvObjects[ResId](
+      mid: MuseumId,
+      dest: StorageNodeId,
+      events: Seq[MoveObject]
+  )(
+      successFailed: Seq[ObjectUUID] => (Seq[ResId], Seq[ResId])
+  )(implicit currUser: AuthenticatedUser, w: Writes[ResId]) = {
+    service.moveObjects(mid, dest, events).map {
+      case MusitSuccess(oids) =>
+        val sf      = successFailed(oids)
```
ingarabr: `val (success, failed)  = successFailed(oids)`

service_storagefacility/app/controllers/storage/StorageController.scala
```diff
@@ -396,82 +410,141 @@ final class StorageController @Inject()(
...
+  private def moveObjectForDelphi(
+      mid: MuseumId,
+      cmd: DelphiMoveCmd
+  )(implicit currUser: AuthenticatedUser): Future[Result] = {
+    val res = for {
+      maybeNode <- MusitResultT(service.getNodeByDatabaseId(mid, cmd.destination))
+      idTuples  <- MusitResultT(objService.getUUIDsFor(cmd.items))
+    } yield {
+      maybeNode.map { node =>
+        val dest = node.nodeId.get // safe...
```
ingarabr: okey, it's safe but the comment doesn't indicate why.

service_storagefacility/app/services/storage/NodeService.scala
```diff
@@ -45,7 +44,10 @@ trait NodeService {
       mid: MuseumId,
       nodeId: StorageNodeDatabaseId,
       envReq: EnvironmentRequirement
-  )(implicit currUsr: AuthenticatedUser): Future[Option[EnvironmentRequirement]] = {
+  )(
+      implicit currUsr: AuthenticatedUser,
+      ec: ExecutionContext
```
ingarabr, line 49: Why do you send around the `ExecutionContext` here and not just import the default from play?

service_storagefacility/app/controllers/storage/StorageController.scala
kpmeen: true enough

service_storagefacility/app/services/storage/NodeService.scala
kpmeen: It's good practice, really. It's a trait that exposes reusable functionality. And it can be used in a context where you want to pass in a different execution context. E.g. when we have the need for an Actor to pull data from the datastore to index, some of these extension services can be useful. But we definitely don't want to use the default play context then. Instead we want to use the actors dispatcher context.
In fact...I'd argue that all our DAO's should have an implicit exec context passed implicitly instead of them importing the default one from play. But that's a big effort in itself.
For this PR it made sense since the code is "new".

service_storagefacility/app/controllers/storage/StorageController.scala
kpmeen: good point

service_storagefacility/app/services/storage/NodeService.scala
ingarabr: But didn't we remove som of them since the import did get a bit messy and since guice will provide the same context when injected ?

musit-service/src/main/resources/reference.conf
```diff
@@ -28,7 +28,7 @@ musit {
   }
 }
 
-ws.proxy {
+play.ws.proxy {
```
ingarabr, line 31: This change will break the proxy. Take a look at the `ProxiedRequest` object

ingarabr (changes requested):

PR #309, opened 2017-05-02
feat: add more sample statuses

ellenjo:
Perhaps this branch will be in conflict with spec-files with MUSARK-904


service_management/test/services/analysis/SampleObjectServiceSpec.scala
```diff
@@ -103,5 +103,36 @@ class SampleObjectServiceSpec
           fail(s"The list contained ${sce.size} elements, expected 1.")
       }
     }
+
+    "successfully add a new sample object with status 'Degraded' " in {
+      val so = SampleObject(
```
kpmeen: Instead of generating the entire object from scratch here. You can add an argument to the `generateSampleObject` function above, that allows you to pass in a `SampleStatus`. If you add it as the last argument and set its default value to `SampleStatuses.Intact`, it should not break any of the other tests.

kpmeen (changes requested):

service_management/test/services/analysis/SampleObjectServiceSpec.scala
ellenjo: I tried it, but it would come out with a message that said that degraded was not an intact...so therefore this solution....

kpmeen (approved):

PR #310, opened 2017-05-03
feature: get treatment list in service_management

ellenjo:
Treatment is in SampleObject. 
Perhaps this branch will be in conflict with musark-904 or  musark-1021


docker/dev/oracle-init.sql
```diff
@@ -438,6 +438,12 @@ CREATE TABLE MUSARK_ANALYSIS.RESULT (
   CONSTRAINT ensure_result_json CHECK (result_json IS JSON)
 );
 
+CREATE TABLE MUSARK_ANALYSIS.TREATMENT (
```
kpmeen, line 442: We should probably have an ID field here to make it easier to identify a specific row. Specially since we're going to have to have an admin API that allows to add/edit these values via an admin client. Doesn't compute well for me to have the language specific columns be the primary key

service_management/app/models/analysis/Treatment.scala
```diff
@@ -0,0 +1,12 @@
+package models.analysis
+
+import play.api.libs.json.{Format, Json}
+
+case class Treatment(
```
kpmeen, line 5: Same applies here as in the table, I believe, it would be useful to have an id field of type `Int`.

service_management/app/controllers/analysis/SampleObjectController.scala
```diff
@@ -105,4 +105,12 @@ class SampleObjectController @Inject()(
 
     }
 
+  def getTreatmentList(mid: MuseumId) =
```
kpmeen: I think this would fit better in a `TreatmentController`. Since we're likely going to add more endpoints for interacting with that type of data (?).

service_management/app/services/analysis/SampleObjectService.scala
```diff
@@ -96,4 +96,7 @@ class SampleObjectService @Inject()(
     soDao.listForMuseum(mid)
   }
 
+  def getTreatmentList: Future[MusitResult[Seq[Treatment]]] = {
```
kpmeen: Same applies here as for the controller and repository implementations. It's good to separate this into a specific `TreatmentService`. It might seem odd to have a service with a single line function. But it helps preserve the modularisation. And makes it easier for some other code that needs access to `Treatment` operations/data to pull it in without having to pull in access to the entire `SampleObject` type.

service_management/test/controllers/analysis/SampleObjectControllerIntegrationSpec.scala
```diff
@@ -188,6 +189,13 @@ class SampleObjectControllerIntegrationSpec
       validateSampleObject(ujs, res.json)
     }
 
+    "list all treatments" in {
```
kpmeen: That also means making separate `Treatment*Spec` implementations for the different test cases

service_management/test/resources/evolutions/test/1.sql
```diff
@@ -76,6 +76,12 @@ CREATE TABLE MUSARK_ANALYSIS.RESULT (
   FOREIGN KEY (event_id) REFERENCES MUSARK_ANALYSIS.EVENT (event_id)
 );
 
+CREATE TABLE MUSARK_ANALYSIS.TREATMENT (
+  no_treatment   VARCHAR2(100)             NOT NULL,
```
kpmeen: See comment on `oracle-init.sql`

service_management/conf/evolutions/default/1.sql
```diff
@@ -77,6 +77,12 @@ CREATE TABLE MUSARK_ANALYSIS.RESULT (
   CONSTRAINT ensure_result_json CHECK (result_json IS JSON)
 );
 
+CREATE TABLE MUSARK_ANALYSIS.TREATMENT (
```
kpmeen, line 81: See comment on `oracle-init.sql`

kpmeen (changes requested):
Generally I'm very pleased with this PR, @ellenjo :-)

There're just a few points I'd like you to address that are more related to packaging of the code related to `Treatment` (i.e. where it "lives").

kpmeen (approved):

PR #311, opened 2017-05-04
feat: add originatedObjectUuid to sample object

ingarabr:
This is a way to easy easier find the originated collection object id.
The reason we need this is that a sample can be taken from another
sample and we will in many cases search for samples taken from the
originated collection object.


kpmeen (approved):

kpmeen:
This should have been a part of PR #308 .

